PR #2, opened 2023-10-02
Feat: Create palette from JSON file (resolves #1)

### LINKED ISSUE #1: Develop Preact components for the palette cells/keys and render them in a grid based on a JSON file
## Palette Cells that Act as Keys
The cells of the palette that are interactive and behave like keys on the palette are defined in a [JSON structure](https://github.com/cindyli/baby-bliss-bot/blob/feat/create-bmw-keys-json/data/bmw_keys.json) given in the [baby-bliss-bot repository](https://github.com/cindyli/baby-bliss-bot/blob/).  (Note:  the links currently reference the branch where the development of this JSON structure is taking place).

The information in the JSON structure is used to:
1. Create a component that is used to render individual cells on-screen, and
2. Lay out the cells according to the row/column arrangement defined in the JSON.

klown:
## Description

**WIP**

This draft pull request uses a [JSON file](https://github.com/cindyli/baby-bliss-bot/blob/feat/bmw/data/bmw.json) to render a BMW palette of encoding symbols as keys.  It defines Preact components for:
- `PaletteCells` that are interactive and behave as keys on an onscreen keyboard,
- `Palette` of such cells, arranged according to the row/column information in the JSON file.

## Related issues
Resolves #1


index.html
```diff
@@ -1,12 +1,17 @@
...
-        <title>Adaptive Palette</title>
-    </head>
-    <body>
-        <div id="app"></div>
-    </body>
+  <head>
+    <meta charset="UTF-8" />
+    <meta name="viewport" content="width=device-width, initial-scale=1.0" />
+    <meta name="color-scheme" content="light dark" />
+    <link href="./src/PaletteCell.css" rel="stylesheet"/>
```
cindyli, line 7: 1. I think it's agreed to use `scss` rather than `css` for styling. Scss is helpful in writing cleaner and more structural code.

2. The styling file doesn't need to be included in the html file. It can be directly imported into its component file. For example, to have `PaletteCell.css` take effect, in the `PaletteCell.ts`, write `import "./PaletteCell.scss"` in the import section.

src/PaletteCell.css
```diff
@@ -0,0 +1,35 @@
...
+ * All rights reserved.
+ *
+ * Licensed under the New BSD license. You may not use this file except in
+ * compliance with this License.
+ *
+ * You may obtain a copy of the License at
+ * https://github.com/inclusive-design/adaptive-palette/blob/main/LICENSE
+ */
+
+.paletteCell {
```
cindyli, line 12: Using scss, line 12-28 can turn into a scoped nested structure:

```
.paletteCell {
  ...

  &:hover, &:focus {
    ...
  }

  & > svg {
    ...
  }
}
```

src/Palette.ts
```diff
@@ -0,0 +1,73 @@
...
+    }
+    bottomRow = aCell.rowStart + aCell.rowSpan;
+    if (bottomRow > rowCount) {
+      rowCount = bottomRow;
+    }
+  });
+  return { numRows: rowCount, numColumns: colCount };
+}
+
+export function Palette (props) {
```
cindyli, line 44: A typescript `type` can be defined to verify the data structure of `props`.

The same comment applies to other component props.

src/Palette.ts
```diff
@@ -0,0 +1,73 @@
...
+  const paletteDefinition = props.json;
+  const rowsCols = countRowsColumns(paletteDefinition);
+  const cellIds = Object.keys(paletteDefinition.cells);
+
+  // Loop to create an array of renderings for each cell
+  const theCells = [];
+  cellIds.forEach((id) => {
+    const aCell = paletteDefinition.cells[id];
+    const paletteCell = html`
+      <${PaletteCell} id="${id}" labelText="${aCell.label}"
```
cindyli, line 55: A palette cell can be one of types: ActionBmwIcon, ContentBmwEncoding etc.

I was thinking the `type` field defined for each cell specifies what type it is, and a component named in this type name can be instantiated to render this cell. If we use this approach, `PaletteCell` component should be renamed to `ActionBmwIcon` (let me know if you come up with a better name), and the `aCell.type` value should replace `${PaletteCell}` in this line.

I'm also thinking each cell type has an `options` block in the JSON for holding the cell specific information. This `options` can be directly when this cell gets instantiated so the instantiator at this line doen't need to worry about what property values are expected. The property structure can be verified inside the cell component. If we do so, the JSON structure for `ActionBmwIcon` will look like:

```
{
    "type": "ActionBmwKey",
    "rowStart": 4,
    "rowSpan": 1,
    "columnStart": 1,
    "columnSpan": 1,
    "options": {
        "label": "ADJ.+ER",
        "BCI-AV-ID": 24879
    }
}
```

Let me know what you think. Thanks.

index.html
klown: Thanks, and good to know.  I'll switch the `css` over to `scss` and `import` into the `.ts` file.

src/PaletteCell.css
klown: Thanks for the suggestion.  I know little about scss nor how it works.  Time to learn :-)

src/Palette.ts
klown: I was going to write an interface for the `props` arguments, and give them a better name, but I was putting that off until the structure was better defined.  So, I will define that interface at some point.

src/Palette.ts
klown: > A palette cell can be one of types: ActionBmwIcon, ContentBmwEncoding etc.

And also "Branch" and "BranchBack" for navigating among different layers.  And others we haven't thought of yet.

> I was thinking the type field defined for each cell specifies what type it is,
> and a component named in this type name can be instantiated to render this
> cell. If we use this approach, PaletteCell component should be renamed to
> ActionBmwIcon (let me know if you come up with a better name), and the
> aCell.type value should replace ${PaletteCell} in this line.

I have been thinking about this, and I keep coming back to using classes since that is a technique for putting common aspects of all palette cells in a base component and deriving more specialized cases therefrom.  I'm not sure how to do that with Functional Preact components, but perhaps there is a way.  Maybe it's as simple as calling "base" Functions from "derived" Functions.  And, in that case, maybe `Function PaletteCell()` is the most basic in the chain.

Another thought I had about the `type` field was to put it into a `data-cellType` attribute in the html itself since all of the action cells are `button` elements.  Without the `data-cellType` attribute, it is impossible to tell what kind of `button` it is (unless the `class` attribute has sufficient information?).  My intuition is that "Undo" or "Delete" are different types of actions than entering a BMW code. But, I don't know yet if there is any use to noting the cell type in the markup.  I'll file away the idea for now.

> I'm also thinking each cell type has an options block in the JSON for holding
> the cell specific information. This options can be directly when this cell
> gets instantiated so  ...

There is a missing verb here:  "... can be directly " ? Copied?  I think you mean something like this in terms of using the `options` block: `<$PaletteCell options=${aCell.options} id=${id} ...`.

Then the `PaletteCell` component, or sub-component, uses only the options that are appropriate for that type.

That's a great idea.  In that case, I am inclined to put the row/column info in the options block as well.

cindyli:
@klown you read my mind and captured exactly what I missed in "can be directly ...". 👍 

I also thought about if a base cell component should be created. The only common information I can think of for now that can be shared by all cells is the row/column information. But I was debating if having a base component handling [this info]((https://github.com/klown/adaptive-palette/blob/feat/create-palette-from-json/src/PaletteCell.ts#L28-L31)) would make it easier or harder to extend from it as the base component doesn't render anything. A better way might be to create a common shared utility component that can be imported when needed, which is a composition rather than an inheritance. What do you think?

This is also why I didn't include row/col values into `options`. Thinking about what you said, it seems making sense to move row/col into `options` as they are cell information and it saves from writing them repeatedly as component props. I will regenerate the BMW palette JSON file with this change.

cindyli:
@klown the new JSON with the `options` block is generated [here](https://github.com/cindyli/baby-bliss-bot/blob/feat/create-bmw-keys-json/data/bmw_palette.json). Note that the filename has been changed to `bmw_palette.json`.

klown:
@cindyli wrote:
> I also thought about if a base cell component should be created. The only common information I can think of for now that can be shared by all cells is the row/column information. But I was debating if having a base component handling this info would make it easier or harder to extend from it as the base component doesn't render anything. A better way might be to create a common shared utility component that can be imported when needed, which is a composition rather than an inheritance. What do you think?

I now think that it's too early to tell what all the cells have in common that constitute a useful base.  For now, I think your earlier suggestion of using the `type` property of a cell to instantiate the appropriate component is a good one and I will try to implement it.  As you know from the discussion in Element, this is not as easy as just doing:  `<${cell.type} ...` because `cell.type` is a string and Preact wants a component function here.  We need to explore a way of mapping the `type` string to the named component.

Then, as we define the components for the palette, if we find a common base, we can do something to handle those common features (I like the idea of using composition to do it, btw).

src/PaletteCell.ts
```diff
@@ -0,0 +1,45 @@
...
+ * compliance with this License.
+ *
+ * You may obtain a copy of the License at
+ * https://github.com/inclusive-design/adaptive-palette/blob/main/LICENSE
+ */
+
+import { html } from "htm/preact";
+import "./PaletteCell.scss";
+
+export function PaletteCell (props) {
```
cindyli, line 15: Regarding the component name, I was thinking it uses the same name as the cell type it supports. The name starts with a more general category performed by the cell. For example, keys that perform actions when pressed are in "action" category, so this cell type and component for BMW code keys are named "ActionBmwCodeKeys"; names for command buttons start with "Command..."; names for display areas start with "Content...". There might be better names than "action", "command" and "content". This naming convention will make files in the `/src` folder easy to navigate since all components in the same category will stay together. What do you think?

src/PaletteCell.ts
```diff
@@ -0,0 +1,45 @@
...
+
+  // Basic styles are the `paletteCell` class defined in PaletteCell.css.
+  // Concatenate any additional classes provided by `props`.
+  let classes = "paletteCell";
+  if (props.class) {
```
cindyli, line 20: There isn't yet `class` and `styles` property. Would it be too early to write code to handle them?

src/Palette.ts
```diff
@@ -0,0 +1,75 @@
...
+  const rowsCols = countRowsColumns(paletteDefinition);
+  const cellIds = Object.keys(paletteDefinition.cells);
+
+  // Loop to create an array of renderings for each cell
+  const theCells = [];
+  cellIds.forEach((id) => {
+    const aCell = paletteDefinition.cells[id];
+    const cellOptions = aCell.options;
+    const paletteCell = html`
+      <${PaletteCell} id="${id}" labelText="${cellOptions.label}"
```
cindyli, line 57: The `cellOptions` can be passed as one single prop so that when this line can be shared for instantiating various components of various cell types:

`<${PaletteCell} id="${id}" options=${cellOptions} />`

index.html
```diff
@@ -1,12 +1,15 @@
...
-    </body>
+  <head>
+    <meta charset="UTF-8" />
+    <meta name="viewport" content="width=device-width, initial-scale=1.0" />
+    <meta name="color-scheme" content="light dark" />
+    <title>Adaptive Palette</title>
+  </head>
+  <body>
+    <h2>Palette Based on JSON</h1>
+    <div id="paletteCell">
```
cindyli, line 11: This div is used to render the entire palette. It might be better to use an id "palette".

klown: The `paletteCell` id was left over from when the code only rendered a single `PaletteCell`.  I have changed it to `bmwKeyCodes` for now, but I would not be surprised if it's changed again.

I also cleaned up `Palette.ts` by removing the rendering code and moving it to `index.html`.

src/Palette.ts
klown: Done.

I've also incorporated your latest `bmw_palette.json` with the correct layout of the BMW code keys.  I noticed that the `type` of key is not within the options block.  I have no opinion if it should be or not, but I'm passing it along like so:

```
<${PaletteCell} id="${id}" options=${cellOptions} type=${aCell.type} />
```

At the moment, `PaletteCell` does nothing with the type, but I presume it will be used when there is a component registry that maps type strings to components.

klown: Replying to myself:  I'm at the wrong level.  If there is a registry, the type will be used by the `Palette` component, something like:

```
<${aCell.type} id="${id}" options=${cellOptions} />
```
There is probably no reason to pass the `type` to the `PaletteCell` component.  I'll take it out at some point.

src/PaletteCell.ts
klown: I like the idea, especially the way that files are grouped within the `/src` folder.  "Command" and "Content" are fine, but "Action" is too generic, I think.  I'm leaning towards "Insert" since the action is to insert content into encoding area and, ultimately, into the message area.

Still, there is nothing within "Insert" to communicate the idea of entering a _sequence_ of codes.  Then again, I'm not sure there needs to be.

src/PaletteCell.ts
klown: I'd say it was too early to remove it :-)

I am using an extra `class` for when the cell is disabled, and this use is shown in the tests, `PaletteCell.test.ts`, like so:

```
<${PaletteCell} id="${id}" options=${cellOptions} class="disabled" />
```

That is handled within `PaletteCell` by:
- setting the `button`'s disabled attribute (`<button disabled ...>`)
- adding the `disabled` class to the `button` so the disabled styles are used ( (`<button disabled class="paletteCell disabled" ...>`

However, having said that, I don't know how disabled is actually handled on the palette.  We know that as bmw codes are entered, certain keys are disabled since they are irrelevant for the current code sequence.  How is that handled by Preact?  This is a case of tracking state and probably involves hooks but I do not know the details.  I will have a look at the documentation.

klown:
@cindyli 

These changes should be enough for your work on the Component registry that uses the `type` field in `bmw_palette.json`:

- There is a share-able `BlissSymbol` component,
- The `PaletteCell` component, its source, test and scss files have all been renamed to `ActionBmwCodeCell`,
- All the code for handling extra classes and styles in the `ActionBmwCodeCell` component has been removed.

The `type` fields in the cells in `bmw_palette.json` have been changed to `ActionBmwCodeCell` as well.  I didn't think to about that until after doing the `git mv ...` locally, committing, and pushing.  At that point I didn't want to go through that process again.  Besides, "cell" is better than "key", right?  😄 
 
However, some of the tests do not run locally for me because of a top-level `await` issue.  I have looked into ways to solve it, but have not succeeded.  I followed the advice in this [github issue](https://github.com/kulshekhar/ts-jest/issues/2057), but I ran into the same problem as [one of the commenters](https://github.com/kulshekhar/ts-jest/issues/2057#issuecomment-1025811045), namely `import` not longer works once you have configured jest and ts as advised.  It has something to do with which version of JS is used, CommonJS vs. ES.  I'm still looking into that.

To be clear:  the code works and the palette renders.  The problem is only with running some of the tests. They fail with the error that top-level await expressions are not allowed.  And that has to do with some settings.

klown:
@hlridge, 

Heads up:  the [build process is complaining](https://github.com/inclusive-design/adaptive-palette/actions/runs/6589654911/job/17904585243?pr=2) that it cannot resolve the `bliss-svg-builder` package:

> [commonjs--resolver] Failed to resolve entry for package "bliss-svg-builder". The package may have incorrect main/module/exports specified in its package.json.

This is how the current project references the builder, on line [15](https://github.com/inclusive-design/adaptive-palette/blob/fd3f0540e60fff4b398bc4ce46c7b7185241970f/package.json#L15) of `package.json`:

```
   "bliss-svg-builder": "github:hlridge/bliss-svg-builder"
```
I'll look into it, but if you see anything obvious, let me know -- thanks.

cindyli:
@klown when running `npm run build` locally, this error is thrown:

```
[vite:esbuild-transpile] Transform failed with 1 error:
assets/index-!~{001}~.js:5772:22: ERROR: Top-level await is not available in the configured target environment ("chrome87", "edge88", "es2020", "firefox78", "safari14" + 3 overrides)
```

It has been fixed in [my PR](https://github.com/klown/adaptive-palette/pull/1) that adds the cell type registry by adding [this line](https://github.com/cindyli/adaptive-palette/blob/feat/add-cellType-registry/vite.config.ts#L15).

klown:
@cindyli:  The first of the latest two commits removes the top-level `await`, replacing it with an `async initAdaptivePaletteGlobals()` function that is called before the rest of the palette machinery is engaged.

This led to a new problem for the tests, where they fail with the message, `ReferenceError: fetch is not defined`, even though the version of node used to run the tests has a version of `fetch()` built-in (v20.8.1).

I followed the advice from a [github issue comment](https://github.com/mswjs/msw/issues/686#issuecomment-1701549602) to set the `globals` object in the jest configruation file (`jest.config.cjs`) to reference the built-in `fetch()`. It works but only partially.

If each test file is run individually there is no problem.  For example:

```
npm run test SvgUtils
```

The above runs the SvgUtils tests and reports actual test errors or success.  The testing framework finds and invokes `fetch()`.

However, the same is not true if all of the tests are run at once:

```
npm run test
```

In that case, `fetch()` is undefined.

I found [this warning](https://jestjs.io/docs/configuration#globals-object) in the jest configuration documentation.

> ... the `globals` object must be json-serializable, so it can't be used to specify global functions. For that, you should use `setupFiles`.

At this point, I am still looking into how `setupFiles` might be used for the case at hand.  It's a list of a set of script files to run before running the tests.  Probably only one setup file needed in our case.  But, it must somehow "import" node's built-in `fetch()` and keep it defined.  This github comment provides a clue:  [https://github.com/mswjs/msw/issues/686#issuecomment-1528832695](https://github.com/mswjs/msw/issues/686#issuecomment-1528832695).  I'll let you know how it goes.

klown:
@cindyli 
I wrote:

> At this point, I am still looking into how setupFiles might be used for the case at hand. It's a list of script files to run before running the tests. Probably only one setup file needed in our case. But, it must somehow "import" node's built-in fetch() and keep it defined. This github comment provides a clue: https://github.com/mswjs/msw/issues/686#issuecomment-1528832695. I'll let you know how it goes.

After exploring github and stackoverflow, I landed on [this comment](https://github.com/jsdom/jsdom/issues/1724#issuecomment-720727999) that states emphatically that one cannot use node's built-in fetch() with jest's js-dom test environment.  Note that we are using js-dom in our tests to get access to dom elements, e.g., the `document` element.

The latest push follows that comment's advice to install the [whatwg-fetch](https://www.npmjs.com/package/whatwg-fetch) module and use the `setupFiles` property to import `fetch` from `whatwg-fetch`.

cindyli:
> After exploring github and stackoverflow, I landed on [this comment](https://github.com/jsdom/jsdom/issues/1724#issuecomment-720727999) that states emphatically that one cannot use node's built-in fetch() with jest's js-dom test environment. Note that we are using js-dom in our tests to get access to dom elements, e.g., the `document` element.

Great to know the root cause. Thanks for figuring it out, @klown.

src/SvgUtils.ts
```diff
@@ -0,0 +1,70 @@
...
+ * argument is an array of BCI-AV-IDs and punctuation, concatenate the array
+ * into a string of builder code strings and punctuation marks.
+ * @param {BciAvId} - The BciAvId to convert.
+ * @return {String} - The concatenation of the builder codes and punctuation,
+ *                    e.g., "B106/B12".
+ */
+export function bciAvIdToString (bciAvId: BciAvId) {
+  let finalCode = "";
+  if (typeof bciAvId === "number") {
+    const { blissSvgBuilderCode } = bciToBlissaryId(bciAvId);
```
cindyli, line 27: When a BCI-AV-ID "1" is given to `bciToBlissaryId()` function, the function doesn't find the corresponding blissary_id and returns `undefined`. It causes this line to fail with error: `TypeError: Cannot destructure property 'blissSvgBuilderCode' of 'bciToBlissaryId(...)' as it is undefined.`.

It would be helpful to add an error handling in `bciToBlissaryId()` to report when a `bci-av-id` cannot find a match. Moreover, if `bciToBlissaryId()` returns `undefined`, the caller function should handle this return without a TypeError.

src/BlissSymbol.ts
```diff
@@ -0,0 +1,42 @@
...
+  // 'TemplateStringsArray.  Property raw is mssing ..."
+  // see: https://stackoverflow.com/questions/50706337/importing-html-to-typescript-to-use-as-templatestringliteral#answer-51012181
+  // Also TypeScript-ESLint does not allow explicit `any` types; override that
+  // rule for this case.
+
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  const templateStringArray = [`${svgMarkupString}`] as any;
+  templateStringArray.raw = [`${svgMarkupString}`];
+
+  return html`
```
cindyli, line 38: I'm wondering if the returned html in a structure below would make more sense semantically?

```
<svg id="{uuid}" />
<lable for="{uuid}">{label-text}</label>
```

What do you think?

klown: There is a bigger issue here @cindyli .

First a `<label>` element cannot used since a `<label>` must reference an `<input>` element.  If there is a need for an accessible name, then either built-in svg  elements, attributes, or text content can be used; or use `aria-label` or `aria-labelledby`. (A short summary of the possibilities is in the [SVG-AAM](https://www.w3.org/TR/svg-aam-1.0/#h-note-9)).

But, the prior question is whether an accessible name is needed.  It is not needed in the case of the `ActionBmwCodeCell` component.  That component is a `button` whose text content serves as the accessible name.   Any other markup within the `<button>`[ is considered "presentational"](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/button_role#all_descendants_are_presentational) or of no semantic significance. What is needed in this case is to block the `<svg>`.  Back when only circles were rendered for the SVG graphics, I added an aria `role="presentation` to the `<svg>` element to make it presentational.  According to the MDN link above, that should not be necessary, but I'll confirm that.  However, if the `BlissSymbol` component automatically added, for example, an `aria-labelledby` to the `<svg>` element, that would negate any presentational role and violate the rule that buttons have no meaningful children.

Now, `BlissSymbol` components are not used only for `ActionBmwCodeCell` components, and the SVG may not be presentational in other contexts.  How about adding a boolean property to the `props` that indicates whether the SVG is presentational or not, such as `isPresentational`?  If the SVG is not presentational, then the <svg> markup could be: `<svg aria-labelledby="{uuid}" .../>`, and the `div` that  currently contains the text label could be `<div><span id="{uuid}">{label}</span></div>`.

Putting it all together and forgetting about the `TemplateStringsArray` coercion, it would like something like:
```
if (isPresentational) {
  return html`
    <svg role="none" />
    <div>label</div>
  `;
else {
  return html`
    <svg aria-labelledby="{uuid}" ... />
    <div><span id="{uuid}">label</span></div>
  `;
}
```

klown: Sure, we can leave it as is until the need for an accessible name arises.  It occurred to me that the `ContentBmwEncoding` and `ContentBmwMessage` areas might need the non-presentational form since the symbols displayed there are not buttons.  What are they?

cindyli: For symbols displayed in the encoding area, I think they are representing the meaning expressed by their labels. For example, if we have a setting in the future that users can choose to only display symbols without displaying their label, those symbols should be announced in the label texts when being read out. My feeling is symbols in the display areas are also meaningful and not just for presentation. But I'm not sure if my feeling is correct.

What do you think, @klown?

klown: From the point of view of a Bliss user, the symbols are never just presentation.  They are everything.  But, from the point of view of accessibility APIs and ATs, they are useless.  Screen readers, for example, cannot speak Bliss at this point in time**.  They can only speak the label text.

@cindyli wrote:
> For symbols displayed in the encoding area, I think they are representing the meaning expressed by their labels.

I think you are correct, so that label text needs to be available to ATs.  If the symbols in the display areas were buttons, then that text would be available.  But they are not buttons and the text has to be exposed in some fashion.  One way is what I outlined above for the non-presentational case.  The question still is, what is their aria role?  The best I can come up with for now is that they are images and I believe SVG elements are images by default (I'll check) .  To be sure, here is a small change to the output for non-presentational `BlissSymbol`:
```
<svg role="img" aria-labelledby="{uuid}" ...  />
<div><span id="{uuid}">label</span></div>
```

> ... if we have a setting in the future that users can choose to only display symbols without displaying their label, those symbols should be announced in the label texts when being read out. 

That's relatively easy by styling the `<span>label</span>` text with `display:none` or perhaps `visibility:hidden`.  Using `@aria-labelledby` will still provide ATs with the label text even if it is not visually rendered.

** - screen readers cannot handle Bliss today, but might in the future.  At one point, emojis were not handled well by screen readers, but that has changed.  Early on the fix was to use `@alt` text or `@arial-label` or `@aria-labelledby`. Now VoiceOver at least has built in speech for many emojis with no need for alt text.  When Bliss is merged into unicode, there will be a standard consistent encoding for each symbol that can be matched to speech output.

klown: Followup re:
>  The question still is, what is their aria role? The best I can come up with for now is that they are images and I believe SVG elements are images by default (I'll check) 

I found a better alternative [`graphics-symbol`](https://w3c.github.io/graphics-aria/#graphics-symbol), but I do not know how well it is supported.  The documentation suggests to set the `role="graphics-symbol img"`.  Browsers use the first role value that they understand so it would default to `img`.

(Aside:  the default role for `svg` elements is _not_ `img` but [`graphic-document`](https://www.w3.org/TR/svg-aam-1.0/#details-id-66).  These symbols are not documents).

cindyli: Your explanation helps me understand better. Your suggestion about adding `isPresentational` flag makes lots of sense. One question to clarify, when Bliss symbols are displayed in the encoding area:

1. If the label is displayed with a symbol, the symbol should be set to "presentation" only;
2. If the label is hidden and only the symbol is displayed, the symbol should be set to "non-presentation".

Is my understanding correct?

klown: After thinking it through, "presentation" vs. "non-presentation" is unrelated to showing or hiding the label.

One issue is the aria role that the `<svg>` element has in different html contexts.  If the `<svg>` element is inside a `<button>` element, it is presentational and has `role="none"`.  That is the case whether the label is visible or not.

If the `<svg>` is in the encoding area, then it is not within a `<button>` element but inside a plain `<div>`.  Its role should be `graphics-symbol img`.  That role is also irrespective of whether the label is visible or not.

Regarding labels, an effective way to handle their visibility _and_ their exposure to ATs is to use `aria-labelledy`.  That's because visibility does not affect `aria-labelledby`.  The accessible name is the same regardless of whether the label text is hidden or not.

Finally, I did a little more research on `<svg>` elements inside `<button>` elements.  There is an issue that `role=none` will not be enough if there is any text inside the `<svg>` element.  If there is text it will be available to ATs even if the main `<svg>` element is marked as presentational; and a screen reader will speak any text inside the `<svg>` tree in that case.  I don't think that the Bliss-SVG-Builder produces anything other than shapes but, to be safe, the recommendation is to add `aria-hidden="true"` to the `<svg>` element here.

Putting it all together, the `BlissSymbol` component should have this logic:

```
if (isPresentational) {
  return html`
    <svg aria-hidden="true" ... />
    <div id="${uuid}">label</div>
  `;
} else {
  return html`
    <svg role="graphic-symbol img" aria-labelledby="${uuid}" > ... />
    <div id="${uuid}">label</div>
  `;
}
```

The ActionBmwCodeCell should change to add the presentational indicator, and an `aria-labelledby`, like so:

```
return html`
  <button id="${props.id}" class="actionBmwCodeCell" style="${gridStyles}" aria-labelledBy="${props.id}">
    <${BlissSymbol} bciAvId=${bciAvId} label=${label} isPresentational="true" uuid="${props.id}" />
  </button>
`;
```

Showing/hiding the label should be handled by SCSS. I have not put a lot of thought into what the selector should be.  Maybe a `blissSymbolLabelText` class or something like that?

cindyli: Thanks for the research and thorough explanation, @klown. Very helpful for me.

I think we probably don't need to worry about the css class for hidding/showing labels in BlissSymbol component for now until the related feature implementation arises.

src/BlissSymbol.test.ts
```diff
@@ -0,0 +1,66 @@
...
+
+  const singleBciAvId = {
+    bciAvId: 12335,
+    label: "VERB"
+  };
+
+  const arrayBciAvId = {
+    bciAvId: [12335, "/", 8499],
+    label: "VERB+S"
+  };
```
cindyli, line 31: It might be clearer to split this test into two sections or two tests, one for testing the number bciAvId and the other for testing the array bciAvId.

setupFetchForJest.ts
```diff
@@ -0,0 +1,19 @@
...
+ * All rights reserved.
+ *
+ * Licensed under the New BSD license. You may not use this file except in
+ * compliance with this License.
+ *
+ * You may obtain a copy of the License at
+ * https://github.com/inclusive-design/adaptive-palette/blob/main/LICENSE
+ */
+
+import { fetch } from "whatwg-fetch";
```
cindyli, line 12: If you happen to have the solution article about using this npm package to fix the use of `fetch()` in the Jest, can you add a code comment to explain what this script is for as well as including the article? Thanks.

src/ActionBmwCodeCell.scss
```diff
@@ -0,0 +1,30 @@
...
+ *
+ * You may obtain a copy of the License at
+ * https://github.com/inclusive-design/adaptive-palette/blob/main/LICENSE
+ */
+
+.actionBmwCodeCell {
+  border: 2px solid blue;
+  background-color: gray;
+  border-radius: 5px;
+  padding: 1em;
```
cindyli, line 16: Can you change scss to use `rem` instead of `em`? The calculation of `em` is based on the size of its parent element while `rem` is always based on the single root element? Thanks.

This comment applies to elsewhere that `em` is used.

src/ActionBmwCodeCell.test.ts
```diff
@@ -0,0 +1,58 @@
...
+  expect(button).toBeValid();
+  expect(button.id).toBe(TEST_CELL_ID);
+  expect(button.getAttribute("class")).toBe("actionBmwCodeCell");
+  expect(button.textContent).toBe(testCell.options.label);
+
+  // Check the grid cell styles.
+  expect(button.style["grid-column"]).toBe("2 / span 1");
+  expect(button.style["grid-row"]).toBe("3 / span 2");
+
+  // Check disabled state (should be enabled)
```
cindyli, line 56: As the `ActionBmwCodeCell` component doesn't handle the disabled state at this point, the check on this attribute probably should be removed. It can be added when the disabled state is handled.

src/ActionBmwCodeCell.scss
```diff
@@ -0,0 +1,30 @@
...
+  font-size: 1em;
+  text-align: center;
+
+  &:hover, &:focus {
+    background-color: lightblue;
+    color: #0000aa;
+  }
+}
+
+.disabled {
```
cindyli, line 26: Same as the previous comment, as the `ActionBmwCodeCell` component doesn't handle the disabled state at this point, the css for it probably should be removed for now.

src/ActionBmwCodeCell.ts
```diff
@@ -0,0 +1,48 @@
...
+ *
+ * You may obtain a copy of the License at
+ * https://github.com/inclusive-design/adaptive-palette/blob/main/LICENSE
+ */
+
+import { html } from "htm/preact";
+import { BlissSymbol, BciAvId } from "./BlissSymbol";
+import "./ActionBmwCodeCell.scss";
+
+function debugProps(x) {
```
cindyli, line 16: Please remove the debugging function before the merge. Thanks.

src/GlobalData.ts
```diff
@@ -0,0 +1,47 @@
...
+import { ActionBmwCodeCell } from "./ActionBmwCodeCell";
+
+export const adaptivePaletteGlobals = {
+  // The map between the BCI-AV IDs and the code consumed by the Bliss SVG
+  // builder.  The map itself is set asynchronously.
+  blissaryIdMapUrl: "https://raw.githubusercontent.com/hlridge/Bliss-Blissary-BCI-ID-Map/main/blissary_to_bci_mapping.json",
+  blissaryIdMap: null
+};
+
+// For debugging
```
cindyli, line 27: Please remove the debugging line before the merge. Thanks.

src/Palette.ts
```diff
@@ -0,0 +1,73 @@
...
+    }
+    bottomRow = cellOptions.rowStart + cellOptions.rowSpan;
+    if (bottomRow > rowCount) {
+      rowCount = bottomRow;
+    }
+  });
+  return { numRows: rowCount, numColumns: colCount };
+}
+
+export function Palette (props) {
```
cindyli, line 44: Please add a test for the `Palette` component.

src/PaletteStore.ts
```diff
@@ -0,0 +1,96 @@
...
+
+  /**
+   * Add a palette to the store, or replace a palette with a new one.  If the
+   * palette's name/identifier matches a palette already in the store, it
+   * replaces it.
+   * @param: {Palette} palette - The palette to add to the store.
+   * @param: {Palette}.name  - The internal name of the palette.
+   * @param: {String} name   - Optional, the preferred name of the palette.
+   */
+  addPalette (palette, paletteName?: string) {
```
cindyli, line 37: Add a type check for `palette` parameter.

src/PaletteStore.ts
```diff
@@ -0,0 +1,96 @@
...
+    PaletteStore.paletteMap[palName] = palette;
+    console.log(`Palette ${palName} added to the store.`);
+  }
+
+  /**
+   * Remove the palette with the given name.
+   * @param: {String} paletteName - The palette to remove.
+   * @return {Palette} reference to the removed palette.
+   */
+  removePalette (paletteName) {
```
cindyli, line 58: Add a type check for `paletteName`. Same comment for line 93.

src/SvgUtils.ts
```diff
@@ -0,0 +1,71 @@
...
+export function getSvgMarkupString (bciAvId: BciAvId) {
+  let builder;
+  try {
+    const svgBuilderArgument = bciAvIdToString(bciAvId);
+    builder = new BlissSVGBuilder(svgBuilderArgument);
+  }
+  catch (err) {
+    console.error(err);
+    console.error(`GETSVGMARKUPSTRING(): using question mark for SVG builder argument from bci-av-id = ${bciAvId}`);
+    builder = new BlissSVGBuilder(DEFAULT_SVG_MARKUP_STRING); // question mark
```
cindyli, line 63: I'm thinking when an error occurs and the SVG builder doesn't return a SVG, would it make more sense to return `undefined` instead of a SVG of a question mark? Returning `undefined` can be captured by the caller component to deal as an exception. Returning a question mark seems like a silent fail that the caller would assume things went well and a question mark is the expected.

src/SvgUtils.test.ts
```diff
@@ -0,0 +1,55 @@
...
+import { initAdaptivePaletteGlobals, adaptivePaletteGlobals } from "./GlobalData";
+import { bciToBlissaryId, bciAvIdToString } from "./SvgUtils";
+
+describe("SvgUtils module", () => {
+
+  // The `singleBciAvId` is taken from the BMW json for "CONJ." The
+  // `bciAvIdArray` is also from the BMW json file using the codes for
+  // "VERB+EN".  The `expectedX` constants are based on a manual lookup of the
+  // blissary ids.
+  const singleBciAvId = 23409;                // CONJ.
```
cindyli, line 22: A more structured way to organize test cases from line 22-26 could be:

```
const testCases = {
  "numberInput": {
    "input": 23409,
    "expected": "B823"
  },
  "arrayInput": {
    "input": [ 12335, "/", 8499 ],
    "expected": "B106/B12"
  },
  "nonexistentBciAvId": {
    "input": 1,
    "expected": undefined
  }
  ...
}
```

The test then loop through `testCases` with the same code to handle each input and check the expected.

src/SvgUtils.ts
```diff
@@ -0,0 +1,71 @@
...
+    bciAvId.forEach((item) => {
+      if (typeof item === "number") {
+        const { blissSvgBuilderCode } = bciToBlissaryId(item);
+        finalCode = `${finalCode}${blissSvgBuilderCode}`;
+      } else {
+        finalCode = `${finalCode}${item}`;
+      }
+    });
+  }
+  return finalCode;
```
cindyli, line 43: When the corresponding Bliss ID is not found for a BCI-AV-ID, it might be better to `console.erro()` the error so the developer knows which BCI-AV-ID is problematic.

cindyli (changes requested):

src/BlissSymbol.test.ts
klown: Okay, done.

docs/DeveloperDoc.md
```diff
@@ -0,0 +1,17 @@
+# Developer Documentation
+
+## How to render a palette
+
+`Palette.ts` constructs a palette based on a JSON file that contains a list
+of the cells in the palette. An example is found in
```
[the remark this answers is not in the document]
klown: The active voice is preferable to the passive. The use of "can be" is passive.  It might be better to say, "An examples is available ...".

docs/DeveloperDoc.md
```diff
@@ -0,0 +1,17 @@
...
+[`src/keyboards/bmw_palette.json`](../src/keyboards/bmw_palette.json). The
+cells object is the list of all of the cells. Each cell has a type `key` and
+an `options` key. The `type` value indicates which component should be used to
+render this cell. The `options` contains information to be passed to the
+component.
+
+When a new `type` value is introduced, developers need to:
```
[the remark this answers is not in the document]
klown: Okay.

package.json
```diff
@@ -8,13 +8,15 @@
     "preview": "vite preview",
     "lint": "run-p lint:*",
     "lint:js": "eslint .",
-    "lint:markdown": "markdownlint-cli2 \"./**/*.md\" \"#node_modules\" \"#dist\"",
+    "lint:markdown": "markdownlint-cli2 \"./**/*.md\" \"#node_modules\" \"#dist\" \"#Bliss-Blissary-BCI-ID-Map\"",
```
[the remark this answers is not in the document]
klown: That was needed when the Blissary Map was a sub-module.  Since we now use `fetch()` to get the map, the sub-module is gone.  So have removed the lint exclusion rule.

setupFetchForJest.ts
klown: Sure.  I have added a comment/link.

src/PaletteStore.ts
klown: That requires a new type.  The only property that the `PaletteStore` uses is the name of the palette, for lookup.  A simple approach for a type then is:

```
type Palette = {
  name: string,
  palette: any
}
```

The problem is that lint will complain about the `any`.  Are we sure of the palette's structure?  If so, there are additional types to replace the `any`:

```
type Palette = {
  name: string,
  palette: JsonPalette  // In future, e.g., JsonPalette | OtherPalette | ...
}

type JsonPalette = {
  name: string,
  cells?: Cell
}

type Cell {
  type: string,
  options: Options,
}

type Options {
  label: string,
  bciAvId: number | (string|number)[],
  rowStart: number,
  rowSpan: number,
  columnStart: number,
  columnSpan: number
}

```

Note:  I'm not sure how to actually specify the `cells` field in the `JsonPalette` type.  It's a variable number of `Cell`s, but it is not an array.

What do think about the two approaches?  Should we use the complete one?

src/PaletteStore.ts
klown: Yes, it is a `string` type.

src/SvgUtils.test.ts
klown: I tried the loop, but have not committed the code.  I found it more difficult to see what is being tested.  I prefer having readable explicit tests to having to think about the parameters in a loop where the tests are executed, and then look up their values in a different block of code.  The loop technique feels opaque to me.

src/SvgUtils.ts
cindyli: I agree this is a design issue. At this development stage, I lean towards the option 1 that does not render any SVG, as well as reporting a `console.error()` with what's happening so developers realize this BCI-AV-ID should be checked.

src/SvgUtils.test.ts
cindyli: Understand. Please go ahead with the approach that works best for you.

src/SvgUtils.ts
klown: There is a `console.error()` where the exception occurs, inside `getSvgString()` (and also in the new `getSvgElement()`).

This might be related:  I don't think `bciToBlissaryId()` should be exported, since it is only used within the SvgUtils module.  The reason that is was exported it was solely for the purpose of writing tests for it.  I have since looked into testing "private" functions and there is more about that in my response below about the addition of tests for the `Palette` component.

If `bciToBlissaryId()` is private, then the only place where the throw will occur is in the two `getSvg...()` functions, where the `console.error()` logs the problem

src/ActionBmwCodeCell.scss
klown: Sure.  I didn't know about `rem` units.  It's good idea.

src/SvgUtils.ts
cindyli: Got it, and great to know the error is already reported.

src/ActionBmwCodeCell.test.ts
klown: I have mixed feelings about removing this.  The test specifically checks if the `button` element is enabled since, as you say, there is no disabled state at this point.  If this test fails, something is very wrong.  When the disabled state is added, then the test will fail, acting as a reminder that there needs to be an additional test for cases where the disabled state is set.

Perhaps we should create an issue about handling the disabled state so we don't forget?

src/ActionBmwCodeCell.ts
klown: Can you clarify?  Do you mean remove just the comment, "`// For debugging`"?  Or, the variable `fetchCount` and the `console.debug()` where the `fetchCount` value is logged?

I have checked the fetch count periodically and only once was it greater than one, which actually led to uncovering a bug.  But since then, it is always been "1".  I'm tempted to remove it at this point.

src/GlobalData.ts
klown: Can you clarify?  Do you mean remove just the comment line, "`// For debugging`"?  Or, the variable `fetchCount` and the `console.debug()` where the `fetchCount` value is logged?

I have checked the fetch count periodically and only once was it greater than one, which actually led to uncovering a bug.  But since then, it is always been "1".  I'm tempted to remove it at this point.

src/ActionBmwCodeCell.test.ts
cindyli: Good point. I understand your concern and agree. Let's leave the check of the disabled state in there then.

src/GlobalData.ts
cindyli: I meant to remove the debugging of the fetchCount completely if the code is stable.

klown:
Hi @cindyli .  I've addressed most of your review.  For the rest I have made comments.  Please give everything another review.  Thanks!

src/ActionBmwCodeCell.ts
```diff
@@ -0,0 +1,47 @@
...
+ * You may obtain a copy of the License at
+ * https://github.com/inclusive-design/adaptive-palette/blob/main/LICENSE
+ */
+
+import { html } from "htm/preact";
+import { BlissSymbol, BciAvId } from "./BlissSymbol";
+import "./ActionBmwCodeCell.scss";
+
+type ActionBmwCodeCellProps = {
+  id: string,
```
cindyli, line 17: "id" -> "type"?

Moreover, the argument passed into this component should only be the `options` part starting at line 18, without the `type` part. The `type` value is only used by `Palette` component to instantiate this component.

src/ActionBmwCodeCell.ts
```diff
@@ -0,0 +1,47 @@
...
+    rowStart: number,
+    rowSpan: number,
+    bciAvId: BciAvId
+  }
+}
+
+export function ActionBmwCodeCell (props: ActionBmwCodeCellProps) {
+
+  const { columnStart, columnSpan, rowStart, rowSpan } = props.options;
+  const { bciAvId, label } = props.options;
```
cindyli, line 31: These two lines can be combined into one line as they are all parsing `props.options`.

src/PaletteStore.ts
klown: I very much like the idea of a common place for shared types; very nice.

I had thought that this technique was not allowed because we were keeping code, styles, etc. grouped by the name of the component.  For example, there is a collection of  `ActionBmwCodeCell.ts`, `ActionBmwCodeCell.scss`, etc. files for everything that has to do with ActionBmwCodeCells.  That made me wonder how to share the types (and interfaces if we ever start to use them).

Breaking out the shared types into a single file makes total sense.  BTW, Where did the name `index.d.ts` come from?  Is that the practice with TypeScript?

cindyli: I've been reading the Preact repo to learn how it organizes the code base. `index.d.ts` is what I learnt there - https://github.com/preactjs/preact/blob/main/src/index.d.ts. Then I searched on `index.d.ts` and found a typescript doc about *.d.ts - https://www.typescriptlang.org/docs/handbook/declaration-files/templates/module-d-ts.html

src/ActionBmwCodeCell.ts
klown: The reason they were split is that putting them all on one line is too long.  I've found a way to split a single destructuring statement across the two lines.

src/index.d.ts
```diff
@@ -0,0 +1,42 @@
...
+// Types associated with the `props` parameter of the component functions.
+
+export type PalettePropsType = {
+  json: JsonPaletteType
+};
+
+export type ActionBmwCodeCellPropsType = {
+  id: string,
+  options: OptionsType
+};
```
cindyli, line 42: These component specific type definitions can be defined in the component as a private type. What do you think?

klown: I just searched the code base, and you are correct.  The only place that  `ActionBmwCodeCellPropsType` is used is inside `ActionBmwCodeCell.ts`.  I'll move that one out of the shared index file and make it local.

Thanks for noticing that!

cindyli: Same with `PalettePropsType`. It's only used in `Palette.ts`.

klown: Done.  I also found that I missed a `BlissSymbolProps` type in terms of the new naming convention.  I fixed that too.

cindyli (approved):

PR #4, opened 2024-01-05
feat: implement a content display area and associated buttons (resolves #2)

cindyli:


.eslintrc.json
```diff
@@ -20,6 +20,7 @@
     "node_modules/"
   ],
   "rules": {
+    "@typescript-eslint/no-explicit-any": ["off"],
```
klown, line 23: Is the rationale that if an _explicit_ `any` is required, then that's allowed?  That the developer would need to consciously and judiciously add the `any` declaration, knowing there was no way around it?  I had to do it once in the case of `BlissSymbol.ts`.  In that case I disabled the check for explicit `any` but just for that one line.  Are there are a lot of places where explicit `any` is used?

Also, why is it not just "off"?  Why the array?

src/CommandClearEncoding.ts
```diff
@@ -0,0 +1,45 @@
...
+  // is to accommodate the component unit test in which the parent palette component is not tested. The
+  // palette state is defined in the palette context.
+  const paletteState = usePaletteState();
+  const setFullEncoding = paletteState?.setFullEncoding;
+
+  const gridStyles = getGridStyle(columnStart, columnSpan, rowStart, rowSpan);
+
+  const cellClicked = () => {
+    setFullEncoding([]);
+  };
```
klown, line 38: Should speak "clear".

src/CommandDelLastEncoding.ts
```diff
@@ -0,0 +1,48 @@
...
+  const paletteState = usePaletteState();
+  const fullEncoding = paletteState?.fullEncoding;
+  const setFullEncoding = paletteState?.setFullEncoding;
+
+  const gridStyles = getGridStyle(columnStart, columnSpan, rowStart, rowSpan);
+
+  const cellClicked = () => {
+    const newEncoding = [...fullEncoding];
+    newEncoding.pop();
+    setFullEncoding(newEncoding);
```
klown, line 40: As above with `ComanndClearEncoding`, should speak "delete".

src/ContentBmwEncoding.test.ts
```diff
@@ -0,0 +1,49 @@
...
+  // Test the content area is rendered properly
+  const encodingAreaByLabel = await screen.findByLabelText("BMW Encoding Area");
+  expect(encodingAreaByLabel.id).toBe(cellId);
+  expect(encodingAreaByLabel.style["grid-column"]).toBe("1 / span 5");
+  expect(encodingAreaByLabel.style["grid-row"]).toBe("2 / span 3");
+
+  // The aria role is defined
+  const encodingAreaByRole = await screen.findByRole("region");
+  expect(encodingAreaByRole).toBeVisible();
+  expect(encodingAreaByRole).toBeValid();
```
klown, line 45: The display area should have a label, and a test that it's properly set.

src/ContentBmwEncoding.ts
```diff
@@ -0,0 +1,43 @@
...
+}
+
+export function ContentBmwEncoding (props: ContentBmwEncodingProps) {
+  const paletteState = usePaletteState();
+  const fullEncoding = paletteState?.fullEncoding ? paletteState?.fullEncoding : [];
+
+  const { id, options } = props;
+  const { columnStart, columnSpan, rowStart, rowSpan } = options;
+
+  // TODO: use a common utility function to calculate the grid position
```
klown, line 30: Use `getGridStyle()` here since it's available, but see my comment elsewhere about the name of `getGridStyle()`.

src/ContentBmwEncoding.ts
```diff
@@ -0,0 +1,43 @@
...
+  const styles = `
+    grid-column: ${columnStart} / span ${columnSpan};
+    grid-row: ${rowStart} / span ${rowSpan};
+  `;
+
+  return html`
+    <div id="${id}" class="bmwEncodingArea" role="region" aria-label="BMW Encoding Area" style="${styles}">
```
klown, line 37: I'm unsure of the use of `@role region` here.  The markup is correct since the region role requires a name or label and a label is provided (but the test does not check for the label). Regions are more like sections of a document than user interface elements, and are [navigation landmarks](https://www.w3.org/TR/wai-aria-1.2/#landmark_roles).  Regions are likely to show up in a table of contents, for example.

The content display area is like a text field, `@role` [`textbox`](https://www.w3.org/TR/wai-aria-1.2/#textbox)` and a read-only text field as well since there is no insertion cursor, selection, nor direct editing of its contents at this point.  It is not focus-able, which is correct.

The accessibility properties of the individual symbols within the display area are not correct, but I"m not sure why yet.  I am investigating.

src/GlobalData.ts
```diff
@@ -9,14 +9,32 @@
  * https://github.com/inclusive-design/adaptive-palette/blob/main/LICENSE
  */
 
-"use strict";
```
klown, line 12: Should we add the [`alwaysStrict` compiler flag](https://www.typescriptlang.org/tsconfig#alwaysStrict) to `ts.config`?  It has the same effect as `"use strict"`, more or less.  I also looked at the stackoverflow article  ["Use Strict" needed in a typescript file](https://stackoverflow.com/questions/31391760/use-strict-needed-in-a-typescript-file) and there are some arguments for making sure the transpiled code has the "use strict"; directive in it, which is part of what `alwaysStrict` does.

src/GlobalUtils.ts
```diff
@@ -0,0 +1,51 @@
...
+
+/**
+ * Generate the grid css.
+ * @param {number} columnStart - The number of the column that the grid item will start from.
+ * @param {number} columnSpan - The number of columns that the item will span across.
+ * @param {number} rowStart - The number of the row that the grid item will start from.
+ * @param {number} rowSpan - The number of rows that the item will span across.
+ * @return {String} - The grid css.
+ */
+function getGridStyle(columnStart: number, columnSpan: number, rowStart: number, rowSpan: number) {
```
klown, line 24: How about `generateGridStyle()` or `makeGridStyle()`?  This does not feel like a `get()` in the sense of retrieving a value.  Note that the documentation for the function is "Generate the grid css".

src/Palette.test.ts
```diff
@@ -77,7 +75,7 @@ describe("Palette component", () => {
   test("Render palette", async() => {
 
     // render() the palette and then wait until its first cell is available to
-    // insure that the entire palette is in the DOM.
+    // ensure that the entire palette is in the DOM.
```
klown, line 78: Thanks for correcting my grammar :-)

src/index.d.ts
```diff
@@ -11,7 +11,7 @@
 
 export type BciAvIdType = number | (string|number)[];
 
-export type OptionsType = {
+export type BlissCellType = {
```
klown, line 14: I'm not sure about "BlissCellType".  In the future the cells will contain symbols other than Bliss but they will behave the same and will have similar options.  The difference will be the `bciAvId` property which will be replaced with some other kind of reference to the actual graphic of the symbol.  How about "SymbolCellType" as a generic name? Or, perhaps "ActionCellType", since the corresponding components are `Action<something>Cell`. Or, perhaps it does need to be more specific to allow different kinds of symbols.  In that case, how about "BlissSymbolCellType".  And if we support, say, the [Widgit](https://www.widgit.com/) symbols, they would be "WidgitSymbolCellType".  What do you think?

src/keyboards/bmw_palette.json
```diff
@@ -1,6 +1,37 @@
 {
     "name": "BMW Palette",
     "cells": {
+        "bmw-encoding-area": {
```
klown, line 4: We should think about making the content display area as a separate JSON and included where necessary.  This part of the display contains the encoding area and the delete and clear buttons.  I merged your changes into my multi-palette branch, and when I switched away from the BMW palette, the content display area and the "Delete" and "Clear" buttons disappeared.  In order to make them render, I had to add their information to the .json files of the other two non-BMW palettes.

This is probably complicated enough to be another issue, and not part of this pull request.  What do you think?

klown (changes requested):
This looks great @cindyli!  I have a few questions and suggestions.

src/ContentBmwEncoding.ts
klown: > The accessibility properties of the individual symbols within the display area are not correct, but I"m not sure why yet. I am investigating.

I figured it out:  the `<$BlissSymbol ...>` needs the property `isPresentation=true`.

.eslintrc.json
cindyli: Good catch, @klown. Allowing the use of "any" type across the project is dangerous. It's better to do it case by case as what you did with `BlissSymbol.ts`. I will remove it. I used "any" when working on this PR but removed them at the end. This line is no longer needed.

src/ContentBmwEncoding.ts
klown: >> The accessibility properties of the individual symbols within the display area are not correct, but I"m not sure why yet. I am investigating.

> I figured it out: the <$BlissSymbol ...> needs the property isPresentation=true.

Actually, perhaps not.  I reverted back to your original code, and noticed that the `aria-labelledby` property for the `<svg>` element was always set to "undefined".  That may explain why a screen reader -- I'm testing with VoiceOver -- "sees" two separate items here.  The first is an unlabelled image and the second is the text.  It's that text that should be referenced by the `aria-labelledby`.  I'm going to experiment with that to see if that solves the problem.

src/ContentBmwEncoding.test.ts
cindyli: Line 37-40 in this test finds the display area by "aria-label", which is to test the area has a properly named label.

src/GlobalUtils.ts
cindyli: Will change to `generateGridStyle()`.

src/GlobalData.ts
cindyli: Good point. I'm thinking to set [the strict flag](https://www.typescriptlang.org/tsconfig#strict) to `true`. This documentation says it enables a wide range of type checking behavior including "alwaysStrict". The linting and running of our code base so far seems working well with `strict: true`.

What do you think?

src/index.d.ts
cindyli: `BlissSymbolCellType` sounds great. Having `Bliss` in the name is because this type structure contains a `bciAvId` field that is specific to Bliss symbols. When we support the graphic symbol or other type of symbols, this field will likely be replaced by the path to the graphic or identifiers to that type of symbol.

src/keyboards/bmw_palette.json
cindyli: Good catch, @klown. We should think about how to make the inclusion of these three elements (the display area, delete and clear buttons) easier.

I agree to address it as a separate issue if needed.

src/GlobalData.ts
cindyli: Setting `strict: true` introduces the type verification in the imported modules. Some of them don't have a proper type definitions. I ended up with what you suggested of `alwaysStrict: true` so that "use strict" is no longer needed in every source file.

src/ContentBmwEncoding.ts
cindyli: As you suggested, using `isPresentation=true` with `BlissSymbol` component fixes the `aria-labelledby="undefined"`. Accessibility properties for symbols in the display area look good now.

cindyli:
@klown Thanks for the review and great suggestions. All have been address. Ready for another look.

klown:
A few of other possible issues that I have noticed while trying it out.

1. If the input area is filled with symbols and one keeps adding to it, then an horizontal scroll bar appears and the whole keyboard starts to narrow.
    - The horizontal scroll bar might be an issue since it adds another user interaction that AAC users don't want to deal with.  I say "might be an issue" since this is a design problem: how do AAC users want the input area to behave?  Should there be a scroll bar, or should it expand to a multi-line text box?  Or something else?  By way of comparison, CBoard behaves the same, adding an horizontal scroll bar when its input area is full.
    - The keyboard narrowing needs to be addressed at some point.  I don't have any recommendations for fixing it.  I don't think it's a blocker, but if you can think of a quick fix, great.
2. I'm getting a lot of Netlify error messages on FireFox when I run the deployed version.  However, I do not see any error messages on Safari.  It's possible that it has something to do with my FF settings.   A message shows up just below the BMW palette that says "Application error: a client-side exception has occurred (see the browser console for more information)."   Here is a sampling of the browser console, in case it means something to you.

```
TypeError: document.getElementsByTagName(...)[0] is undefined        vendor.bundle.js:9246:26
DOMException: The operation is insecure.                             bugsnag.js:2457:26
DOMException: The operation is insecure.                             bugsnag.js:2457:15
A client-side exception has occurred, see here for more info: https://nextjs.org/docs/messages/client-side-exception-occurred
                                                                     bugsnag.js:2457:15
DOMException: The operation is insecure.                             bugsnag.js:2457:15
Eight more similar DOMExcptions
```

cindyli:
@klown thanks for these helpful findings. All addressed and ready for another review.

klown (changes requested):
There are a few small tedious things to fix and once fixed, I can't think of any reason to not merge this.

src/ContentBmwEncoding.scss
```diff
@@ -0,0 +1,30 @@
+/*
+ * Copyright 2023 Inclusive Design Research Centre, OCAD University
```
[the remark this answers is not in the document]
cindyli: Being picky in code review is a valuable trait. 👍

cindyli:
@klown new comments have been addressed. Ready for another look. Thanks.

src/Palette.ts
```diff
@@ -22,11 +22,12 @@ type PalettePropsType = {
  * Given a palette defined in a json structure, compute the number of rows
  * and columns in that palette.
  *
- * @param {Object} paletteDefinition - An object that lists the positions,
- *                 heights and widths of the cells in the palette.
+ * @param {Object} paletteDefinition - An object in the pre-defined type of
```
klown, line 25: I have a thought.  Would it be clearer to say `{JsonPaletteType}` instead of `{Object}`?

I admit that I have not been consistent, but I did use `{BciAvIdType}` as one of the `@param`s and `{BlissSVGBuilder}` as the return type in SvgUtlis.ts, see the documentation for the [`getSvgBuilder()`](https://github.com/inclusive-design/adaptive-palette/blob/2132549cc65c4debd79b8efec08165cdefddd333/src/SvgUtils.ts#L47) function.  I did that consistently with SvgUtils.ts, at least.  I did not follow that pattern at all in PaletteStore.ts (unfortunately).

We had as similar approach in GPII where we specified `@param {Component}` to show that the parameter was a fluid component, and a return type of `{Promise}`, which were actually `fluid.promise` objects.

What do you think?  Even if you agree, I don't think we should use this PR to go around and fix all of the documentation.  We go either do it as we edit the files, or make it a separate small task to fix those details.

cindyli: I see your point and fixed the type here.

klown (approved):
Looks good to me, @cindyli!

Issue #10, opened 2024-04-11
Separate the output area from any palette in the display.


klown:
## Is your feature request related to a problem?
The output area, specifically, the `ContentsBmwEncoding` component, is a region where the symbols that the user has chosen appear as they are selected.  Currently, this area is duplicated at the top of each description file containing a palette of symbols.  As such, the output area is an inherent part of each palette, and they are rendered together as a single palette.

## Describe the solution you'd like
In terms of the palette description files, the output area should be in its own separate file.  With respect to the display, it should be a separate region on the page, independent of any other component.  Finally, as users choose symbols, its contents should be updated as appropriate.

klown:
Closed #10 as completed via 15f5c1d.

PR #11, opened 2024-04-11
feat: separate output area from symbol palettes

klown:
Code changes for separating the output display area of symbols from other palettes.  In brief:

- Move the JSON description of the `ContentsBmwEncoding` component to its own `.json` file,
- Remove the repetition of the same description from each of the other palette `.json` files,
- Create a `<div>` for rendering the `ContentsBmwEncoding` component on the page,
- Respond to symbol selections on various palettes by updating the contents of the `ContentsBmwEncoding` component,
- Respond to the "Clear" and "Delete" symbols.

Addresses issue #10

### DIFF
--- a/src/palettes/palette_file_map.json
+++ b/src/palettes/palette_file_map.json
@@ -3,5 +3,6 @@
   "My Family Palette": "myfamily",
   "People": "people",
   "BMW Palette": "bmw_palette",
-  "Palettes": "palettes"
+  "Palettes": "palettes",
+  "InputArea": "input_area"
 }
--- a/index.html
+++ b/index.html
@@ -8,13 +8,13 @@
   </head>
   <body>
     <h2>Palettes Based on JSON</h1>
-    <div id="outputArea">
-      <div id="output_palette"></div>
-    </div>
     <div id="backupArea">
       <div id="backup_palette"></div>
     </div>
-    <div id="mainPaletteDisplayArea">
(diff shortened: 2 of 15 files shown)


src/palettes/palette_file_map.json
```diff
@@ -3,5 +3,6 @@
   "My Family Palette": "myfamily",
   "People": "people",
   "BMW Palette": "bmw_palette",
-  "Palettes": "palettes"
+  "Palettes": "palettes",
+  "InputArea": "input_area"
```
cindyli, line 7: The use of "InputArea" and "input_area" don't match with "outputArea" and "output_area.json" used elsewhere. Fortunately, the output area is only rendered once at the instantiation so this bug doesn't cause any problem. :)

It seems reasonable either calling this area an input or an output. I personally lean towards "input" as it's an area to receive user inputs. :)

klown: Good point, @cindyli .  If this was not bliss, it would a standard text _input_ field for entering text.  I'll switch it around to consistently use "input" in the code and json.  But I won't rename the branch nor the issue.

cindyli: > But I won't rename the branch nor the issue.

I agree.

klown: Fixed.  Ready for another review, thanks.

cindyli (approved):

PR #12, opened 2024-05-21
Demo: Interact with Ollama service runinng locally

klown:
1. Add a `demos` folder to the project for various demonstrations.
2. Add the two demo Ollama chat web-apps that run on the `locahost` dev server to the `demos` folder.
3. Add documentation for all of the above.
   - Documentation about the `demos` folder in the main README,
   - Documentation on how to set up and run the Ollama chat demo.


demos/Ollama Chat Service/README.md
```diff
@@ -0,0 +1,95 @@
...
+Ollama's [Model
+Library](https://github.com/ollama/ollama?tab=readme-ov-file#model-library) to
+install and run that language model.
+
+Once the Ollama Service is installed, launched and running, proceed to the next
+steps.
+
+## Chat Web-App
+
+There are two versions of the web-app.  They appear identical within the browser
```
cindyli, line 22: It would be helpful to compare or explain the pros and cons of these two versions.

demos/Ollama Chat Service/ollama.js
```diff
@@ -0,0 +1,61 @@
...
+  else {
+    executeAsk();
+  }
+}
+
+// Function for passing the chat prompt using ollama service.
+async function queryChat (query) {
+  const message = { role: "user", content: query };
+  const response = await ollama.chat({
+    model: "phi3",
```
cindyli, line 31: It might be more adaptable to define the model as a variable at the beginning of the script. The "phi-3" portion in the heading `<h1>Experimenting with Ollama (phi-3) in a Browser</h1>` could be adjusted based on the model value. Moreover, update the README to describe where to adjust this variable.

This will make it easier for trying different models.

demos/Ollama Chat Service/whatExpress.html
```diff
@@ -0,0 +1,97 @@
...
+<script>
+
+  const ollamaService = "http://localhost:11434";
+
+  // Handle click on "Ask" button
+  function askClicked(theButton) {
+    const single = "Answer with a single grammatically correct sentence."
+    console.debug("Button clicked");
+    // Empty out the response area
+    document.getElementById("phi3Output").innerText = "Working...";
```
cindyli, line 13: If the script is more flexible with any Ollama models, it would be better not having "phi3" in names such as "phi3Output".

cindyli (changes requested):

demos/Ollama Chat Service/whatExpress.html
klown: Thanks for catching that, @cindyli.  I made a similar change to the `ollama.html` web-app and its `ollama.js` script at some point, but didn't make the parallel change to `whatExpress.html`.

demos/Ollama Chat Service/README.md
klown: Done.

klown:
I've addressed all of you comments @cindyli -- ready for another review.

cindyli (changes requested):
It would be helpful to use different h1 titles for two demos. For example, one title is suffixed with "(using fetch() in browser)" while the other one's title is suffixed with "(using ollama library)". It will help to tell which demo is being used.

cindyli:
One issue I noticed is, when nothing is entered in the "Prompt" field, clicking either of two buttons will also give responses that talk about random things. Since waiting for a response takes a little while, shall we add a check at button clicks to remind users to enter texts in the prompt if the prompt value is null?

klown:
Good points @cindyli .

Regarding the headings the `README.md`, I have changed two of them to call out how the web-app works, using `fetch()` vs. using the ollama client library.

With respect to actual text in the prompt field, I decided to disable the "Ask" buttons until there is text to use for the query with and there is an LLM to query.

Ready for another review.

demos/Ollama Chat Service/ollama.html
```diff
@@ -0,0 +1,26 @@
+<html>
+<head>
+  <title>Experimenting with Ollama in a Browser</title>
+</head>
+<body>
+  <script type="module">
+    import "./ollama.js";
+  </script>
+  <h1>Experimenting with Ollama in a Browser</h1>
```
cindyli, line 9: Thanks for adjusting the demo titles in README. Can you also adjust <h1> and <title> texts in actual demos? When running any of these demos, it's helpful to know which version it is.

klown: Yes, that makes sense.  And, I think the <h1> should match the new <title> as well.  I guess that is what you were pointing out, and not the headings in the `README.md`.

Okay, fixed.

PR #14, opened 2024-06-17
fix: improve text label for the prompt text entry area

klown:
This is a simple PR to adjust the label on the text area for creating the input to send to the LLM.  Instead of the uninformative "Prompt:" , it is now "What does this express: ?"


klown:
This should not be much work @cindyli, but just so there are another pair of eyes looking at it -- thanks.

cindyli (approved):

PR #17, opened 2024-07-11
feat: add an initial Express server code

cindyli:
At today's IDRC tech check-in, it was suggested that we consolidate the "Adaptive Palette" and "BBB server" into a single repository since they are interdependent. This would streamline the sharing of common features and code, such as shared types, npm tasks etc. I think this is a great idea. Additionally, we can replace the Vite server with our own Express server.

In this pull request, the "src/" directory is now organized into two sub-directories: "client/" for the client-side rendering code, and "server/" for the Express server. The "src/server/server.js" file is configured to serve the client HTML from the "/dist" directory.

What are your thoughts on this approach?

@agamba, I couldn't find your GitHub ID in the reviewer list. Could you please take a look when you have some time? Thanks!


klown (commented):
Here are my first impressions.  This is a brainstorm about what the possible implications are and questions about them.

How difficult is it to continue to use cloudflare with the new approach, including deploying development versions?  I suspect it's easy, but are there any gotchas?  We ran into issues with vite/rollup where localhost development worked, but the production version did not and it was hard to debug and track down the issue.  This is actually a question about the implications of using Express with local development, external server development, and production, using Preact and the deployment system.

The adaptive palette code depends very much on the [Bliss-Blissary-BCI-ID-Map service](https://github.com/hlridge/bliss-blissary-bci-id-map), but no one is suggesting that we move its source into this repository, nor move the adaptive palette source into the Bliss-Blissary-BCI-ID-Map service repository.  And that is because as long as the http end points stay the same, how the code in the mapping service develops is independent of the internal workings of the palette client.  To state the obvious, that's the nature of any client/server relationship --  generally, their internal workings are independent implying the code bases are independent too.  However, @cindyli  noted that putting it all in once place promotes "... the sharing of common features and code, such as shared types  ..."  That's a really good point.  Still, there are questions:

- The adaptive palette client is written in TypeScript and defines supporting types and interfaces.  What services will be written in TypeScript and use the existing types/interfaces?  Whisper? Ollama? The prompt generator? Others?
- Some of the services are written in python.  How easy is it to use TypeScript types with python?  Or, are the python based services going to be ported to TypeScript/express?
- What is it like use TypeScript to make a service?  In fact, this pull request answers that.  I need to take a closer look at it.
- I know of one project, [Clusive](https://github.com/cast-org/clusive/tree/development), that was primarily a service, but also provided a substantial client-side front-end.  Most of the client-side code was part of same repository which made it easy to integrate changes to server side with changes to the front-end.  Testing, debugging, and deploying was easy because "everything was in the same place".  However, there was one front-end piece that had its own separate repository: the [R2D2BC](https://github.com/d-i-t-a/R2D2BC) EPub book reader.  Furthermore, that made sense since it was developed as a module to be used anywhere that a web-based EPUB reader could be used, not just as part of Clusive.

I hope I don't sound negative here; I'm thinking out loud to explore the implications to this change. And, I am definitely warming to the idea of having it all in one place, while wondering to what degree some parts actually are independent.

Thoughts about my thoughts?

cindyli:
@klown, thank you for sharing your thoughts and concerns. These are excellent discussion points.

1. Regarding the cloudflare deployments, as I understand it, Cloudflare is primarily for deploying static websites, not for sites that require a hosting web server. We can continue using Cloudflare to deploy the client side of this project. However, for deploying both the client and the server, we'll need a cloud service such as AWS or Azure etc.

2. Regarding the types in TypeScript that can be shared between the server and client code, one obvious example is the API data structures exchanged between the client and server. We could organize our type definitions into three scripts: one for shared types, one for client-specific types, and one for server-specific types.

3. The server we plan to develop is meant to be consumed by the client chatbot. We do not envision this server as an individual service like Ollama or WhisperLive. Instead, it will function as an integrated component within a specific design sequence. Additionally, if the client and server are hosted on different servers, the workflow will involve extra steps to start two servers, configurations to connect two servers, extra Docker images, and so on.

I'm in the process of applying TypeScript to the server code. Looks promising so far.

cindyli:
I missed one point you mention about how easy to apply TypeScript to Python. Python is much strong typed than Javascript. Reinforcing python with TypeScript may not be necessary. It would be nice to do more research in this part.

klown:
> I missed one point you mention about how easy to apply TypeScript to Python. Python is much strong typed than Javascript. Reinforcing python with TypeScript may not be necessary. It would be nice to do more research in this part.

Here is one data point, again from Clusive, that argues for having strong types on the client side.

Some of the activities that the user engaged in were considered to be worth persisting.  One of these was a set of timings on a given page.  There was client side javascript for capturing the timings, placing them in a POJ, and putting that on a messagequeue (aside:  the message queue was a fluid component).  The messagequeue used ajax to POST the timing information to Clusive's "/messagequeue/"  endpoint.  On the python side, the timing event was a python object, more specifically, a python model object (a row within a table in the database).  It's very strongly typed.

But the client side was just vanilla javascript with no types.  That led to tricky debugging in that you would have to (1) look at the POJ sent from the client side using the browser's debugger to make sure the fields were filled in properly, and with the right types.  Then, (2) in the python debugger, where the python end point handler was called, double check that the data in the request was the right type for the python object, or it would fail.  Which is a long way to agree with you that setting up the correct types in the client and linting the code would be a very good pre-flight check that the right data is sent to the server.  It is desirable that the Types and Interfaces match what the python request handlers are coded to handle.

cindyli:
> Here is one data point, again from Clusive, that argues for having strong types on the client side.

I totally agree with applying TypeScript to Javascript. In our case would be both the client side and the server side.

src/server/server.ts
```diff
@@ -0,0 +1,32 @@
...
+
+// Parse incoming requests with JSON payloads
+app.use(express.json());
+
+// Serve static files from the root dist directory
+app.use(serveStatic(path.join(__dirname, "../../../client")));
+
+app.listen(PORT, () => {
+  console.log(`Server is running on http://localhost:${PORT}`);
+});
```
klown, line 32: What do you think about a `/health` end-point?  For now, all it would do is respond "healthy".  But as the app is developed and other things are included, e.g., a database, the `/health` end-point could report whether all those extra connections were running.  For example:

```
app.get(`/health`, (req, res) => {
 const health = { status: "healthy" };
 health.ollamaService = // check that Ollama is available;
 health.database = // check any database requirements.
 ...
 res.json(health);
}
```

package.json
```diff
@@ -3,39 +3,50 @@
...
-    "dev": "vite",
-    "build": "vite build",
-    "preview": "vite preview",
+    "start": "run-s build:* start:server",
+    "build:client": "vite build",
+    "build:server": "tsc --project tsconfig.server.json",
+    "start:server": "node dist/server/src/server/server.js",
+    "dev": "run-p dev:* start:devServer",
+    "dev:watchClient": "vite build --watch",
+    "dev:watchServer": "tsc --project tsconfig.server.json --watch",
```
klown, line 12: This not a comment on the code, but just documenting a technique I used to test the "watch" mode for the vite build , tsc, and nodemon.  I used the `touch` shell command to change the date of various files and folders to see what triggered a rebuild and restart.  For the record:

First, the `dist` directory tree:

- `touch dist/client/assets/index-DuP7TeSo.js` caused `nodemon` to restart.
- `touch dist/client/assets/index-DuP7TeSo.css` or the `.map` did not cause `nodemon` to restart
- `touch dist/client/assts/index-DuP7TeSo.js` caused `nodemon` to restart.
- `touch dist/server/config/config.js` caused `nodemon` to restart.
- `touch dist/server/src/server/server.js` caused `nodemon` to restart.
- `touch dist/client` did nothing.
- `touch dist/client/assets` did nothing.
- `touch dist/server/config/` did nothing.
- `touch dist/server/src/` did nothing.

Next, using the same technique in the `src` directory tree:

- `touch src` did nothing.
- `touch src/client` did nothing.
- `touch src/client/ActionBmwCodeCell.scss` caused `vite build` to run. In  turn, this caused `nodemon` to restart (multiple times).
- `touch src/server/server.ts` **did nothing**.

The last one was surprising.  In order to trigger the `tsc` script even when in "watch" mode, it was necessary to change something in the file and save it.  That is different from the `vite build` script which does react to a change in the file's date.

klown (commented):
Except for the changes I suggested in [my PR](https://github.com/cindyli/adaptive-palette/pull/1), this is fine as far as I'm concerned.

package.json
klown, line 12: I also have run into an issue where changing a palette does not cause a re-build by `vite build --watch`. While it does trigger `nodemon` to restart, the new palette in `./pubic/palettes/some_palette.json` is not copied to `./dist/client/palettes/some_palette.json`.  The server needs to be manually stopped and restarted for the palette changes to be copied over to the `dist` folder.

I tried adding a `build.watch` property to `vite.config.ts`, like so, but it did nothing:
```
build: {
 watch: {
  include: "public/palettes/*.json"
  }
...
}
```

The [documentation for the `watch` property](https://vitejs.dev/guide/build#rebuild-on-files-changes) suggests looking at the [rollup watch configuration](https://rollupjs.org/configuration-options/#watch) for more options, which is where the idea to include `./public/palettes/*.json` came from.  However, that is entirely for files involved in rollup, and not static files.  So far, I have found no way to watch for changes in static files.  This [github issue](https://github.com/vitejs/vite/issues/4970) made a request for this feature, but the [conclusion](https://github.com/vitejs/vite/issues/4970#issuecomment-1500960824) was that it wasn't needed.

There is also this [stackoverflow](https://stackoverflow.com/questions/69626090/how-to-watch-public-directory-in-vite-project-for-hot-reload) article, but I have not had the time to do a deep dive into it.

package.json
```diff
@@ -3,39 +3,50 @@
...
   "type": "module",
   "scripts": {
-    "dev": "vite",
-    "build": "vite build",
-    "preview": "vite preview",
+    "start": "run-s build:* start:server",
+    "build:client": "vite build",
+    "build:server": "tsc --project tsconfig.server.json",
+    "start:server": "node dist/server/src/server/server.js",
+    "dev": "run-p dev:* start:devServer",
```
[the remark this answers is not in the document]
cindyli: I modified the `start:devServer` task to wait until the `dist/server/src/server/server.js` is built ready then executing `nodemon` on it: `"start:devServer": "wait-on dist/server/src/server/server.js && nodemon dist/server/src/server/server.js"`. Although `nodemon` still restarts a couple of times when `server.js` is in the process of building, the number of the restarting is significantly reduced.

package.json
```diff
@@ -3,39 +3,50 @@
...
+    "dev:watchClient": "vite build --watch",
+    "dev:watchServer": "tsc --project tsconfig.server.json --watch",
+    "start:devServer": "nodemon dist/server/src/server/server.js",
```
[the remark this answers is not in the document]
cindyli: Thanks for catching it and issuing a pull request to fix. Your PR has been merged.

src/server/server.ts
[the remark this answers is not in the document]
cindyli: I didn't find a nice and clean way to report these values too. Let's re-visit when we need to.

cindyli: I added a `/health` endpoint that returns a JSON of `{ status: "healthy" }` for now.

package.json
cindyli: I also looked into this issue and followed the [stackoverflow solution](https://stackoverflow.com/questions/69626090/how-to-watch-public-directory-in-vite-project-for-hot-reload) you found to write a Vite plugin:

```
import { PluginOption } from "vite";

export default function deployPublicDir(): PluginOption {
  return {
    name: "public-to-dist",
    handleHotUpdate({ file, server }) {
      console.log("in handleHotUpdate: ", file);
      if (file.includes("palettes")) {
        server.ws.send({
          type: "custom",
          event: "public-file-updated",
        });
      }
    },
  };
};
```

This problem is the `handleHotUpdate()` option is never triggered when a file in `public` is updated. This seems because the `public` directory is never watched in the Vite watch mode:

* [Vite watch documentation](https://vitejs.dev/guide/build#rebuild-on-files-changes) says: `With the --watch flag enabled, changes to the vite.config.js, as well as any files to be bundled, will trigger a rebuild.` Changes to files to be bundled will trigger a rebuild, but files in `public` are not bundled via rollup. They are just copied to the destination root in the building process.
* The conclusion comment in the Vite [github issue](https://github.com/vitejs/vite/issues/4970) also says: `The conceptual model is that files in public are not watched for HMR during dev either.`

So, I'm not sure how the stackoverflow solution managed to get `handleHotUpdate()` triggered with a change to the `public` directory. I wonder if `i18n` package did something special in this case, or if the `public` directory in that example is a part of the Vite bundle?

I think for now we will need to manually restart the building process when there's a change to `public` directory.

PR #24, opened 2024-09-30
feat: add strategies palette to make use of vocabulary strategies for finding symbols

klown:
**WIP**

This adds a "Strategies" palette and its sub-layers for selecting among various ways to help the user find a specific Bliss symbol.

A new type of cell is introduced here, namely the `LabelCell` (a Preact component).  Its purpose is to label the palette that it is part of.  In terms of HTML, the component is rendered as a `<label>` element, but it is unlikely that this right way to provide the label.

To Do:
- [ ] Complete the strategies functionality.
  - [ ] Create a new type of cell, similar to a toggle button, to allow the user to set the strategy mode when they reach the strategy they want to use.  Currently the leaf strategy cells are `ActionBmwCodeCell` components, but they should cause the strategy to "work". 
- [ ] Address the labeling relationship between the `LabelCell` and the palette it labels, e.g., by adding an `aria-labelledby` or `aria-label` to the appropriate place,
- [ ] Improve the styling of the `LabelCell`,

Note that this PR includes the changes from #23.

This PR replaces [a previous PR](https://github.com/cindyli/adaptive-palette/pull/3) that was not based on the `main` branch.  The code has been appropriately re-based and this new PR issued.


cindyli:
Closed as this feature is no longer requested.

PR #27, opened 2024-12-19
Feat: add composition information

klown:
This captures and stores the composition information for every Bliss symbol in the authorized vocabular.  The  composition information is available in `./data/bliss_symbol_explanations.json` in case someone wants access to it.

1. This uses a TypeScript script and `npx vite-node` to execute it.
2. The code removes the `composingIds` arrays from `bliss_symbol_explanations.json` since they are often different from the new`composition` array.
3. Some of the BCI AV IDs do not have any composition information defined for them -- there are three.  The webapp lists them, and the information is available in `BlissSymbolComposition.md` with the `./docs` folder.  Is there a better place for this information?
4. Also, there are slightly over 1000 BCI AVI IDs where the composition information is identical to the symbol's BCI AV ID.  In that case, no `composition` array is stored with the symbol.  These are also listed by the webapp and in `BlissSymbolComposition.md`


klown:
Hi @cindyli :  This is almost ready for review, but see my comments in the main description above.  I've left it as a draft PR.  If you have any use for the composition information, however, it's available as described above.

cindyli:
@klown, I agree using a script to update `./data/bliss_symbol_explanations.json` is a good idea. 

Since this script is likely to be run only once to update the file, and the composition information is not expected to change frequently, manual updates might be enough for future changes. When the composition information for the three currently missing symbols becomes available, a manual update might be an easy route too. With that in mind, here are some thoughts:

1. Change this webapp script to a node.js script and save it in the `./scripts` folder.
2. Write a documentation explaining the purpose of the script, its usage, and its one-time nature. It should only be used again if there is a big update to the composition information.
3. Remove `./data/BciAvCompositionAnalysis.tsv` from the repository. Instead, include instructions in the documentation on how to generate this intermediate file from its source spreadsheet, for example, saving the spreadsheet as a `.tsv` file.

Let me know what you think. Thanks.

cindyli (approved):

PR #28, opened 2025-01-24
Feat: add an indicator palette to allow users to add indicators to the last symbol in the input area

klown:
This PR is for handling the placement of indicator symbols (e.g., plural, action) to the last symbol in the input area.

Code changes:
- adds a new action cell type: `ActionIndicatorCell`,
- `GlobalData.ts` is initialized by importing the information in `bliss_symbol_explanations.json`,
- adds a new type for explicitly pairing a symbol's BCI-AV-ID and its composition array,
- `SvgUtils.ts` has information about indicator and modifier IDs and functions for handling them within an SVG builder string.

UI change:
- a preliminary indicators palette for demonstrating adding indicators to an existing symbol


src/client/SvgUtils.ts
```diff
@@ -20,6 +20,29 @@ type BlissaryMapEntryType = {
   blissSvgBuilderCode: string
 };
 
+// Ranges and list for all the indicator symbols.  The range values are the
+// minimum and maximum BCI AV ID.
+const indicatorIds = {
```
cindyli, line 25: There are many indicators. Can you explain why these indicators are selected and grouped in this way? What do the terms "rangeN" and "list" mean?

Some indicators such as "25458" and "28046" are marked as "OLD" in the Blissary dictionary, which means they are no longer in use. Why are they still selected?

Can you add a code comment following each array to explain what these indicators represent in English to improve the readability. Thanks.

src/client/ActionIndicatorCell.ts
```diff
@@ -0,0 +1,99 @@
+/*
+ * Copyright 2024 Inclusive Design Research Centre, OCAD University
```
cindyli, line 2: The years in copyright should be updated to 2024-2025 now. :)

src/client/ActionIndicatorCell.ts
```diff
@@ -0,0 +1,99 @@
...
+      // If there are no indicators on the symbol, then place the indicator
+      // above the first symbol that is not a modifier.  Otherwise, replace the
+      // current indicator with the new one at the same position.
+      // 1. `classifierIndex` is the index of the classifier in the array,
+      // 2. the next index is the separator between the classifier and the next
+      //    symbol, e.g., "/": `classifierIndex+1`,
+      // 5. insert the ";" separator for indicators followed by the indicator id,
+      // 7. insert the rest of the array as it was.
+      if (indicatorPositions.length === 0) {
+        console.log(`newBciAvId: ${newBciAvId}`);
```
cindyli, line 54: There are a number of `console.log()` and `console.debug()` calls in the code. Can you go through to remove? Thanks.

src/client/ActionIndicatorCell.ts
```diff
@@ -0,0 +1,99 @@
...
+        newBciAvId = [
+          ...newBciAvId.slice(0, classifierIndex+1),
+          ";", indicatorBciAvId,
+          ...newBciAvId.slice(classifierIndex+1)
+        ];
+      }
+      indicatorPositions.forEach((position) => {
+        // NOTE: This will not work if a single BCI AV ID represent a Bliss-word
```
cindyli, line 62: To solve the issue you described in this note, do you think if the composition JSON file should be updated so single Bliss words are decomposed to the level with "characters + indicators"?

src/client/ActionIndicatorCell.ts
```diff
@@ -0,0 +1,99 @@
...
+    }
+    // The BCI AV ID is a single identifier, not an svg builder array.
+    // Note: cannot handle this situation in general unless: the symbol is a
+    // Bliss-character.
+    else {
+      newBciAvId = [ newBciAvId, ";", indicatorBciAvId ];
+    }
+
+    const payload = {
+      // TODO:  what should the following two fields be?  For now the ID is
```
cindyli, line 79: Can we discuss this problem with Shirley and others in the language meeting - What kind of announcement Bliss users are expected or would be helpful to them? There are a number of announcements I can think of:

1. Announce the symbol label and the indicator label, as what the palette does now.
2. Announce the word with its form, for example "persons" instead of "person". For other odd forms like the adverb form of "run", a LLM request can be used to figure out the best word for it, which might have nothing to do with "run" anymore.
3. Support multiple announcement methods and allow users to control it by a setting.

src/client/GlobalData.ts
```diff
@@ -12,14 +12,16 @@
 /**
  * Populate and export global data
  */
-
 import { signal } from "@preact/signals";
 
+import bliss_symbols from "../../public/data/bliss_symbol_explanations.json";
```
cindyli, line 17: As we discussed yesterday, this line leads to a warning message. Can you have a look on how easy or hard to fix it? Although it's a warning, I'm worried once we live with it now, if it causes a failure in the future, we need to re-investigate this whole thing again. Let me know what you think. Thanks.

src/client/SvgUtils.ts
cindyli: Resolved in a discussion. The decision is to use ranges as is.

src/client/ActionIndicatorCell.ts
klown: I ran into a similar problem in the branch I am currently working on, with a word that is not in the authorized vocabulary.  The word was "don't know" made from "know" + "not" -- I showed you the issue at last Monday's meeting.  A solution is to decompose down to Bliss-characters.

For example, in terms of an SVG builder string, "don't know" is "15161/15733", where 15161 is "know".  If you look it up in the blissary, you will see that it has an action modifier, which is not evident from its ID 15161.  However, my new `decomposeBciAvId()` turns  "15161/15733" into "15162;8993/15474/14947", and the indicator (8993) is now easily found.  That will solve the current problem as well.

The question is: should I add that change to this PR, or leave it as part of the next iteration?  I think it belongs here -- what do you think?

src/client/ActionIndicatorCell.ts
klown: Slavery!  :-)

Should I do this for all of the files, or just the ones that are part of this PR.  That is, only the ones that have been modified?

src/client/ActionIndicatorCell.ts
klown: Good catch, thanks.  I will get rid of them.

src/client/ActionIndicatorCell.ts
cindyli: I agree it belongs here.

src/client/ActionIndicatorCell.ts
cindyli: It might be easier to do a find and replace across all files. If you think only files touched in this PR should be updated, I'm fine too.

src/client/ActionIndicatorCell.ts
klown: I recall that Susie made a comment a while back during one of my technical updates.  She said something like, "Current AAC devices know how to say the plural or the past tense, but saying 'child, plural' or 'think, past tense' is not terrible".  I think I replied that, yes, this was for demonstration only, and that we needed designers to tell us what should actually happen.

I would happy with any UX advice here.  I *think* that modern  AAC devices do provide the correct grammatical form, but they do not use an LLM for that.  However, I don't know what they use; we should consult with Susie.

The "TODO" comment also relates to how to handle the `id` field of the `payload`.  In the past, it was the symbol's BCI AV ID, but that does not apply here.  For the time being it's a simple concatenation of the original symbol's ID plus the indicator's ID.  Maybe it should be the SVG builder string?  I'll give it some more thought.

src/client/GlobalData.ts
klown: Sure at some point.  One fact to note is that this warning only happens when serving up one of the demo apps using the vite server.  It does not happen using express.  The first thing I'd look into is whether serving the demo apps with express would eliminate the warning.

src/client/ActionIndicatorCell.ts
klown: I manually changed some of the copyright dates of the files in this PR as I made code changes to that file. Then I manually fixed the rest (10 files) in one go, but only those in this PR.  I'm not sure what the pattern is to look for in order to automate this.  I left the rest alone.

src/client/ActionIndicatorCell.ts
klown: I got rid of these and some others.  I hope I got them all.

src/client/ActionIndicatorCell.ts
klown: Okay, it's now part of this PR, as well as the improvements to the Palette Generator app.

src/client/GlobalData.ts
klown: As we discussed, I added a comment about the warning from the `vite` server here.

klown:
Okay @cindyli, this is ready for another review -- thanks!

cindyli (changes requested):
If I click a indicator cell without selecting a Bliss word first, this error is reported in the console log: `ActionIndicatorCell.ts:39 Uncaught TypeError: Cannot read properties of undefined (reading 'bciAvId') at cellClicked (ActionIndicatorCell.ts:39:33)`.

Indicator cells probably should be disabled when the input area is empty.

apps/palette-generator/index.html
```diff
@@ -121,19 +123,24 @@ <h1>Palette Generator</h1>
...
       <li><code>SVG:13166,";",9011:SVGLABEL:children</code> will generate the
-          an SVG image that combines the symbols that mean "child" and "plural".
-          The cell's label is "children".
-            <ul>
-              <li>
-                  If no <code>LABEL:</code>is provided, the cell will have no
-                  label.
-              </li>
-            </ul>
+          SVG image that combines the symbols that mean "child" and "plural".
```
cindyli, line 132: This explanation and the initial examples in the text area are very useful. Do you think if the use of ";" and "/" should also be explained? Such as, ";" is to superimpose the second symbol on the first one, while "/" is to have them displayed side by side.

Moreover, it would be nice to have one or more examples in the text area to show the usage of both.

cindyli (commented):
The second row of the indicator palette have from "connection" to "sounds like". Are they modifiers that should be used together with one or more Bliss words? For example, the "opposite" should be place before a Bliss word to make sense. The current implementation places these modifiers as standalone Bliss words.

The solutions off the top of my head are:
1. Discuss proper ways to display and use them;
2. I understand those modifiers are probably from Andius' example palette. But if we cannot implement their usage properly, it might be better not to include them in the indicator palette for now.

klown:
> If I click a indicator cell without selecting a Bliss word first, this error is reported in the console log: `ActionIndicatorCell.ts:39 Uncaught TypeError: Cannot read properties of undefined (reading 'bciAvId') at cellClicked (ActionIndicatorCell.ts:39:33)`.
> 
> Indicator cells probably should be disabled when the input area is empty.

Indicator action cells are now disabled when there is nothing to "indicate", and they are styled with a grey backgroun (CSS silver)

klown:
> The second row of the indicator palette have from "connection" to "sounds like". Are they modifiers that should be used together with one or more Bliss words? For example, the "opposite" should be place before a Bliss word to make sense. The current implementation places these modifiers as standalone Bliss words.
> 
> The solutions off the top of my head are:
> 
>     1. Discuss proper ways to display and use them;
> 
>     2. I understand those modifiers are probably from Andius' example palette. But if we cannot implement their usage properly, it might be better not to include them in the indicator palette for now.

You are right, this due the work with Andrius' suggestions where several of his keyboards have the strip across the top with these modifiers, and because I merged that branch with this one.  But, since I do not know how these actually function, I'm taking them out for now; but I'll continue the work in the other branch

apps/palette-generator/index.html
klown: Thanks for that suggestion.  I've added more examples and found a way, although awkward, of placing Bliss symbols into the HTML.

klown:
Thanks for your suggestions @cindyli  -- all addressed.

PR #33, opened 2025-06-26
fix: provide full decomposition of bliss-words in palette json files

klown:
This tweaks some Bliss-words in the static palette files to use their full decomposition instead of partial decompositions or just single BCI AV IDs.


klown:
@cindyli Based on our earlier element chat, I've gone through the palettes and found a few more cases where the full decomposition was not done during palette creation.  Could you look at the changes and see if you see any goofs -- this is tedious and error prone.

I would not be surprised if there other cases that I have missed, but it's difficult to come up with a good technique to find them all.

PR #41, opened 2025-12-18
feat: keyboard support for input area

jobara:
- moves `faiss-node` to be an optional dependency
- refactors cursor movement so that the functionality can be used by the buttons and key presses
- adds a `clamp` function
- adds arrow key cursor navigation to the input area

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,7 +13,7 @@ To work on the project, you need to install [NodeJS and NPM](https://nodejs.org/
 for your operating system.
 
 _**Note:** If you'd like to make use of RAG (optional), you'll also need to ensure that [CMake](http://cmake.org)
-is installed. CMake is required for installing `faiss` which is pulled in by the optional depenency `faiss-node`._
+is installed. CMake is required for installing `faiss` which is pulled in by the dependency `faiss-node`._
 
 Then, clone the project from GitHub. [Create a fork](https://help.github.com/en/github/getting-started-with-github/fork-a-repo)
 with your GitHub account, then enter the following in your command line
@@ -29,10 +29,6 @@ to install dependencies:
 ```bash
 npm ci
 ```
-
-_**Note:** If CMake is not available on the device, the optional dependency `faiss-node` will fail to install but
-will not prevent the application from installing. However, if you wish to use RAG, you'll need to install CMake and
-re-run `npm ci`._
 
 ## Development
 
@@ -99,9 +95,9 @@ Follow these steps to complete a one-time setup to enable RAG in the system:
 
 #### Troubleshooting
 
-In order to use RAG, `fais-node` must be installed. `faise-node` is an optional dependency and requires that
-[CMake](https://cmake.org/) be installed on the machine first. If this has not been done, install CMake and re-run
-the application install steps.
+In order to use RAG, `fais-node` must be installed. `faise-node` requires that [CMake](https://cmake.org/)
+be installed on the machine first. If this has not been done, install CMake and re-run the application install
+steps.
 
 ### Lint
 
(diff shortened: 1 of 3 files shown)


src/client/ContentBmwEncoding.ts
```diff
@@ -66,6 +66,36 @@ function generateMarkupArray (payloadArray: Array<EncodingType>, caretPos: numbe
   });
 }
 
+export function moveCursor (positionChange = 1) {
```
klown: Provide the type of `positionChange` (number).

src/client/GlobalUtils.ts
```diff
@@ -102,9 +102,26 @@ function insertWordAtCaret (wordToAdd: SymbolEncodingType, symbolSet: SymbolEnco
...
+ *  - `min` if `value` was less than `min`
+ *  - `max` if the `value` was greater than `max`
+ *  - `value` if it fell within the `min` `max` range.
+ *
+ * @param {number} value - The value to evaluate
+ * @param {number} min - The minimum value to be returned
+ * @param {number} max - The maximum value to be returned
+ * @returns {number} - The constrained value
+ */
+function clamp (value: number, min: number, max: number) {
```
klown: Just a note -- no change necessary.  I thought that a `clamp()` function might already exist as a built-in within TypeScript or JavaScript, and asked duck.ai if there was one.  Its reply was that there was none, and also provided code.  It was the same as yours, but without the type declarations.  So, your implementation must be right :-)

src/client/ContentBmwEncoding.ts
```diff
@@ -66,6 +68,60 @@ function generateMarkupArray (payloadArray: Array<EncodingType>, caretPos: numbe
...
+  moveCursor(-1);
+}
+
+export function moveCursorToHome () {
+  moveCursor(Number.NEGATIVE_INFINITY);
+};
+
+export function moveCursorToEnd () {
+  moveCursor(Number.POSITIVE_INFINITY);
+};
```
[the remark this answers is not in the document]
jobara: Also I'm using `INFINITY` here so that I don't have to make calculations based on `changeEncodingContents.value`, but that would be another option.

klown:
Thanks @jobara.  I have run into an issue and am trying to decide how to handle it.  Here's the background:

I added a [global key handler](https://github.com/jobara/adaptive-palette/blob/d3c5a09d630eed04a2bd762525ab38f02523c7ca/src/client/index.js#L46) for navigating back to a previous palette, with the idea that users could access it at any time, no matter what was focused.  However, the actual keystroke is the back-tick (`) which _can_ be a legitimate character in a text input.  Thus, there is [code to check](https://github.com/jobara/adaptive-palette/blob/d3c5a09d630eed04a2bd762525ab38f02523c7ca/src/client/index.js#L57) whether keyboard focus is in any kind of text input control so that the back-tick can be added to that text.  Since the symbol input area has a role of "textbox", once it has focus, the "navigate back" keystroke stops working.  It probably shouldn't.

I can see a fix where if focus is in the symbol input area, then a back-tick keystroke is not ignored.  I could make a branch from your branch and add the one or two lines to accomplish that.  Or you could look into it.

What do you think?

klown:
@jobara I found another issue.  When the "Backward" and "Forward" buttons are pressed, the cursor moves backward or forward, _and_ there is a voice output, "backward" or "forward", respectively.  When the arrows keys are used to move the cursor, it is silent.  This is inconsistent.

klown:
@jobara  a followup to one of my earlier comments:
> ...
> I can see a fix where if focus is in the symbol input area, then a back-tick keystroke is not ignored. I could make a branch from your branch and add the one or two lines to accomplish that. Or you could look into it.
> 
> What do you think?

I've created the branch and made the fix.  Should I go ahead and make a PR for you to review?

klown (changes requested):
LGTM -- thanks for adding those keystrokes, @jobara!

As discussed, you can now factor out all the non-keyboard control aspects of this PR, the rag/faiss/make issue.

PR #48, opened 2026-02-05
docs: update CMake install info and add info about OpenMP install.

### LINKED ISSUE #47: Feature: Add information about requirements for installing faiss-node

There is a trouble shooting note that CMake needs to be installed first. However I also found that I needed OpenMP installed as well.

Under the troubleshooting section that mentions CMake is needed, mention installing OpenMP. I used homebrew to install cmake and libomp. 

### Describe alternative solutions you've considered

_No response_

### Additional context or notes

_No response_

jobara:
* [x] This pull request has been tested by running `npm run test` without errors
* [x] This pull request has been built by running `npm run build` without errors

## Description

Updates the troubleshooting section with following:

- fixes typo in `faiss-node` name
- adds info about installing CMake from Homebrew
- adds info about needing OpenMP

## Steps to test

1. Review troubleshooting section under RAG

### DIFF
--- a/README.md
+++ b/README.md
@@ -95,6 +95,8 @@ Follow these steps to complete a one-time setup to enable RAG in the system:
 
 #### Troubleshooting
 
+##### CMake not installed
+
 In order to use RAG, `faiss-node` must be installed. `faiss-node` requires that [CMake](https://cmake.org/)
 be installed on the machine first. If this has not been done, install CMake and re-run the application install
 steps.
@@ -105,13 +107,31 @@ Using homebrew, you can install CMake with the following:
 brew install cmake
 ```
 
-If you encounter an `'omp.h' file not found` error, it likely means that you need to also install
+For other installation options please consult [CMake's download page](https://cmake.org/download/).
+
+##### 'omp.h' file not found
+
+If you encounter an `'omp.h' file not found` error, it likely means that you need to install
 [OpenMP](https://openmp.llvm.org/).
 
-Using homebrew, you can install CMake with the following:
+Using homebrew, you can install OpenMP with the following:
 
 ```bash
 brew install libomp
+```
+
+##### npm error ENOTEMPTY: directory not empty
+
+If you encounter an `npm error ENOTEMPTY: directory not empty` error, you can try removing the `node_modules`
+directory and rerunning the npm install.
+
+```bash
+
+# remove the node_modules directory
+rm -rf node_modules
+
+# re-run the npm install
+npm ci
 ```
 
 ### Lint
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -95,10 +95,25 @@ Follow these steps to complete a one-time setup to enable RAG in the system:
 
 #### Troubleshooting
 
-In order to use RAG, `fais-node` must be installed. `faise-node` requires that [CMake](https://cmake.org/)
+In order to use RAG, `faiss-node` must be installed. `faiss-node` requires that [CMake](https://cmake.org/)
 be installed on the machine first. If this has not been done, install CMake and re-run the application install
 steps.
 
+Using homebrew, you can install CMake with the following:
```
klown, line 104: How about: "Using homebrew on macOS, you can install ..."

Also, what do you think of adding something like the following for those who are not using a Mac?

"Please consult [CMake's download page](https://cmake.org/download/) to see how to install CMake on other platforms or OSes."

README.md
```diff
@@ -95,10 +95,25 @@ Follow these steps to complete a one-time setup to enable RAG in the system:
...
+
+```bash
+brew install cmake
+```
+
+If you encounter an `'omp.h' file not found` error, it likely means that you need to also install
+[OpenMP](https://openmp.llvm.org/).
+
+Using homebrew, you can install CMake with the following:
```
klown: I think you mean "... you can install **opm.h** with ..." or something like that, not CMake, right?

README.md
```diff
@@ -95,10 +95,25 @@ Follow these steps to complete a one-time setup to enable RAG in the system:
...
+
+```bash
+brew install libomp
+```
+
```
klown, line 122: As it happened, I tried out the above step-by-step.  That is, first I installed CMake using homebrew, and then ran `npm i` immediately after. That resulted in the "omp.h" error.

Then, step two, I used homebrew to install libomp.  Immediately after that was done, I tried `npm i` again.  That led to an error about a non-empty directory:

```
npm error ENOTEMPTY: directory not empty, rename '.../adaptive-palette/node_modules/faiss-node' -> '.../adaptive-palette/node_modules/.faiss-node-WCBpMouc'
```
I think what happened was faiss-node was partially installed at step 1 before running into the "omp.h" error and when the process was repeated, it warned of the existing faiss-node directory, and stopped.

In an attempt to fix that problem, I deleted the `node_modules` folder.  That solved it, `npm i` worked.

I think we need to add a suitable warning to this README -- what do you think?

klown (changes requested):
Hi @jobara.

I have a couple of suggestions and one correction -- see my comments.  Otherwise, this is fine.

jobara:
> Hi @jobara.
> 
> I have a couple of suggestions and one correction -- see my comments. Otherwise, this is fine.

@klown I've updated based on your suggestions.

klown (approved):
LGTM

PR #54, opened 2026-03-26
fix: upgrade bliss-svg-builder to rc.1

hlridge:
## Summary
- Upgrades `bliss-svg-builder` from `^0.1.0-alpha.10` to `^1.0.0-rc.1`
- No application code changes. The rc.1 API is backward-compatible for the constructor, `.svgCode`, and `.svgElement` usage in this project
- Adds a `structuredClone` polyfill to the Jest setup file: `bliss-svg-builder@rc.1` uses `structuredClone` internally, which Node.js has had since v17 but Jest's jsdom environment does not expose, the same class of issue that already required a `fetch` polyfill in `setupFetchForJest.ts`

### DIFF
--- a/setupFetchForJest.ts
+++ b/setupFetchForJest.ts
@@ -1,5 +1,5 @@
 /**
- * Copyright 2023-2024 Inclusive Design Research Centre, OCAD University
+ * Copyright 2023-2026 Inclusive Design Research Centre, OCAD University
  * All rights reserved.
  *
  * Licensed under the New BSD license. You may not use this file except in
@@ -19,8 +19,9 @@ import { fetch } from "whatwg-fetch";
 
 // jsdom does not expose structuredClone even though Node.js has it since v17.
 // bliss-svg-builder@1.0.0-rc.1 depends on it.  Same class of issue as fetch above.
+
+import v8 from "node:v8";
 if (typeof globalThis.structuredClone === "undefined") {
-  const v8 = require("node:v8");
   globalThis.structuredClone = (val: unknown) => v8.deserialize(v8.serialize(val));
 }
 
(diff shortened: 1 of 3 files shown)


setupFetchForJest.ts
```diff
@@ -17,6 +17,13 @@
 
 import { fetch } from "whatwg-fetch";
 
+// jsdom does not expose structuredClone even though Node.js has it since v17.
+// bliss-svg-builder@1.0.0-rc.1 depends on it.  Same class of issue as fetch above.
+if (typeof globalThis.structuredClone === "undefined") {
+  const v8 = require("node:v8");
+  globalThis.structuredClone = (val: unknown) => v8.deserialize(v8.serialize(val));
+}
```
klown, line 26:
```suggestion
// jsdom does not expose structuredClone even though Node.js has it since v17.
// bliss-svg-builder@1.0.0-rc.1 depends on it.  Same class of issue as fetch above.

import v8 from "node:v8";
if (typeof globalThis.structuredClone === "undefined") {
  globalThis.structuredClone = (val: unknown) => v8.deserialize(v8.serialize(val));
}
```
The use of `require()` causes a lint error.  You can see the error if you execute `npm run lint`.  However this is easily fixed by using `import` as shown above.

setupFetchForJest.ts
klown:
```suggestion
 * Copyright 2023-2026 Inclusive Design Research Centre, OCAD University
```

klown (changes requested):
Thanks for this fix @hlridge!  I ran the tests yesterday afternoon after upgrading to node v24.14.1 but the same errors occurred.  I looked more closely at them and saw a common error message: `ReferenceError: structuredClone is not defined`.  Then I noticed your PR and started to look into it.  It fixes the problem.  Still, there are a couple of minor changes -- see my comments.

Thanks again.

klown (approved):
LGTM.

PR #57, opened 2026-03-31
Feats: Find by gloss plus ollama integration

klown:
* [x] This pull request has been tested by running `npm test` without errors
* [x] This pull request has been built by running `npm run build` without errors

## Description

This pull request contains a merge of all the work from two other pull requests: PR #44 (feat/find-by-gloss-CLONE) and PR #42 (feat/integrate-ollama-api-with-palette).  The problem is that merging those into the main branch individually will be complicated and have numerous conflicts.  This branch/pull-request has done all of the conflict resolution, and can be merged directly into main.  That is, you may prefer to merge using this combined PR rather than the individual PRs to avoid the messy conflicts.

That said, as of this writing this PR is in sync with the current state of PRs 44 and 42.  If anything is changed in either one of those, then this will be out-of-date.


cindyli:
@jobara, @chosww - this pull request is ready for your reviews.

src/client/ActionSvgEntryField.ts
```diff
@@ -0,0 +1,121 @@
...
+    
+    speak(payload.label);
+    setMalformed(false);
+    form.reset(); // Clear the form for the next entry
+  };
+
+  return html`
+    <form onSubmit=${svgToSymbol} class="actionSvgEntryField">
+      <fieldset>
+        <legend>Enter symbol using SVG builder string</legend>
```
chosww, line 94: Are we going to localize the palette in the future? If so, it is a good idea to store user facing strings in an object file and import them.

cindyli: Localization should certainly be on the TODO list. I agree with your idea.

PR #58, opened 2026-04-06
feat: compose words

klown:
* [x] This pull request has been tested by running `npm run lint` without errors
* [x] This pull request has been tested by running npm run typecheck without errors
* [x] This pull request has been tested by running `npm test` without errors
* [x] This pull request has been built by running npm start without errors

## Description

This is an attempt at providing the means for users to combine symbols from their palettes to compose Bliss-words; users can use this feature to generate a symbol with indicator and modifier, or optionally mark them with the combine markers and transfer them to the existent symbol input field.

This branch is based on https://github.com/inclusive-design/adaptive-palette/pull/135.

For the code review, please use [the compare URL](https://github.com/cindyli/adaptive-palette/compare/chore/remove-rag-support...klown:adaptive-palette:feat/compose-words) to view the specific changes that need to be reviewed.

## Steps to test

1. Set the "compose" switch to "on".
2. Symbols selected from the main palette area are placed in the compose word input area,
3. The symbols can be navigated with the move-caret buttons,
4. The indicator and modifier strip symbols can be added or removed from the symbols in the compose word area,
5. The combine marker toggle associated with the compose word area can be used to surround the composition with the combine marker,
6. The "delete" button will delete the symbol at the caret,
7. The "clear all" button will clear all of the symbols from the compose word input area,
8. The "add" button will transfer the composed symbol to the main input area and flip the "compose" switch. to off  At this point the palette returns to normal input and edit mode.

Note: once the composed symbol has been added to the main input area, it is treated as a unit, and can no longer be edited.

## Additional information

The UI needs work and there are several new Preact components.  These include:

- `ContentComposeWordsEntry`:  similar to the existing `ContentBmwEncoding` component.
- `ToggleIsComposingWords`: switch to allow user to change between normal operation and composing words mode.
- `ToggleMakeCombination`: button for adding/removing the combine marker to the composed word.
- `CommandAddCompostion`: button to transfer the composed word to the main input area and set the compose switch to "off". 

Also, currently the label of composed words just appends labels of words and there should be a way to predict meaning of the compose words, which will be addressed in the future work.


src/client/GlobalData.ts
```diff
@@ -130,6 +127,34 @@ export const changeEncodingContents = signal<ContentSignalDataType>({
   caretPosition: -1,
 });
 
+/**
+ * Signals for tracking the contents of the word composition input area when
+ * the user is composing a Bliss-word.  The contents signal has the same
+ * structure as the `changeEncodingContents` above.  The isComposing is a state
```
cindyli: The code comment for `isComposing ` is here but the actual constant is defined at line 156. Please re-organize.

src/client/GlobalData.ts
```diff
@@ -130,6 +127,34 @@ export const changeEncodingContents = signal<ContentSignalDataType>({
...
+export const composeWordContents = signal<ContentSignalDataType>({
+  payloads: [],
+  caretPosition: -1,
+});
+
+/**
+ * Some of the CommandXxx components use an `aria-controls` to associate them
+ * with the element that they control.  For thes components, there is an
+ * `ariaControls` field in the associated palette definition .json file.  The
+ * follwoing map defines which content signal (see immediatly above) goes with
```
cindyli: typo: follwoing -> following

src/client/GlobalUtils.ts
```diff
@@ -132,10 +135,127 @@ function clamp (value: number, min: number, max: number) {
...
+
+/**
+ * Remove the last symbol at the caret the given contents.
+ * Note: what is really wanted is something that removes the last
+ * Bliss-character.
+ *
+ * @param {ContentSignalDataType} encodingContents - Contents to add the symbol to.
+ * @return {ContentSignalDataType} - the modified contents.
+ */
+function removeLastSymbol (encodingContents: ContentSignalDataType): ContentSignalDataType {
```
cindyli: The entire content of this function is commented out. This function doesn't seem to be used anywhere. If so, it can be removed. Please double check. Thanks.

src/client/CommandAddComposition.ts
```diff
@@ -0,0 +1,74 @@
...
+ */
+
+import { v4 as uuidv4 } from "uuid";
+import { VNode } from "preact";
+import { html } from "htm/preact";
+import { BlissSymbol } from "./BlissSymbol";
+import { changeEncodingContents, composeWordContents, isComposing } from "./GlobalData";
+import { BlissSymbolInfoType, LayoutInfoType } from "./index.d";
+import { insertWordAtCaret, generateGridStyle, speak } from "./GlobalUtils";
+import "./ActionIndicatorCell.scss";
```
cindyli: This is `CommandAddComposition` component but it imports `actionIndicatorCell.scss`. Also, the component doesn't seem to be using any css in the imported file. Please check if this line should be removed.

src/client/index.js
```diff
@@ -100,10 +110,12 @@ function getRequiredElement(id) {
  */
 function elementAllowsTextEntry(element) {
   if (!(element instanceof HTMLElement)) { return false; }
-  return element.id !== INPUT_AREA_ID && (
+  const isInputArea = element.id === INPUT_AREA_ID;
+  const isComposeArea = element.id === COMPOSE_AREA_ID;
+  return !isInputArea || (isComposeArea && (
```
cindyli: This function is used to allow users to use backtick to trigger the "go back" function when the focused element is not in the input area. The logic before the change is to return true when the given element is a text area.

The new logic is opposite: when the focused element is in the input area, the function returns false. The user experience becomes when the user is in the input area, the backtick triggers go back. When outside of the input area, the backtick will stop working. It breaks the original intention.

This function probably should not be changes.

src/client/GlobalUtils.ts
```diff
@@ -132,10 +135,127 @@ function clamp (value: number, min: number, max: number) {
   return Math.min(Math.max(value, min), max);
 }
 
+/**
+ * Add the given symbol to the current contents at the given caret position.
```
cindyli, line 139: This function is only used for adding modifier. Should the function be rename to addModifier? `isModifier` flag may not be needed too.

src/client/ActionModifierCellCommon.ts
```diff
@@ -13,10 +14,12 @@ import { VNode } from "preact";
 import { html } from "htm/preact";
 import { BlissSymbolInfoType, LayoutInfoType } from "./index.d";
 import { BlissSymbol } from "./BlissSymbol";
-import { changeEncodingContents } from "./GlobalData";
-import { generateGridStyle, speak } from "./GlobalUtils";
+import { INPUT_AREA_ID, COMPOSE_AREA_ID, contentSignalMap, isComposing } from "./GlobalData";
+import { composeBlissWord, generateGridStyle, speak } from "./GlobalUtils";
 import "./ActionModifierCell.scss";
 
+const ISA_MODIFIER = true;
```
cindyli: This might be a typo for "ISA_MODIFIER". It can be removed if the `composeBlissWord()` function is only for adding modifier.

src/client/ContentEncodingInputField.ts
```diff
@@ -0,0 +1,146 @@
...
+      return html`
+        <div class="blissSymbol">
+          <${BlissSymbol} composition=${payload.composition} label=${payload.label} isPresentation="true" />
+        </div>
+      `;
+    }
+  });
+}
+
+export function incrementCursor (contentSignal: Signal<ContentSignalDataType>) {
```
cindyli, line 67: Shall cursor functions from line 74-122 be moved to `CursorActions` component?

src/client/ContentEncodingInputField.ts
```diff
@@ -0,0 +1,146 @@
+/*
+ * Copyright 2023-2026 Inclusive Design Research Centre, OCAD University
```
cindyli: Copyright headers should no longer include years. Please follow the format used in `GlobalUtils.ts` as the reference.

This comment applies anywhere else years are used.

src/client/ToggleMakeCombination.ts
```diff
@@ -0,0 +1,80 @@
...
+  "composition": 233,
+  "modifierInfo": []
+};
+
+type ToggleMakeCombinationPropsType = {
+  id: string,
+  options: BlissSymbolInfoType & LayoutInfoType
+};
+
+export function ToggleMakeCombination (props: ToggleMakeCombinationPropsType): VNode {
```
cindyli: Would `ToggleCombineMarker` be a better name for this component?

src/client/ToggleMakeCombination.ts
```diff
@@ -0,0 +1,80 @@
...
+  const cellClicked = () => {
+    if (payloads.length === 0) {
+      return;
+    }
+
+    if (isCombinedNow) {
+      const updatedPayloads = uncombineContent(composeWordContents.value, SymbolCompositionToSkip);
+      if (updatedPayloads !== composeWordContents.value) {
+        composeWordContents.value = updatedPayloads;
+        speak("remove combination");
```
cindyli: remove combination -> remove combine markers

src/client/ToggleMakeCombination.ts
```diff
@@ -0,0 +1,80 @@
...
+      }
+    } else {
+      composeWordContents.value = combineContent(
+        composeWordContents.value, COMBINE_MARKER_PAYLOAD
+      );
+      speak("add combination");
```
cindyli: add combination -> add combine markers

src/client/ToggleMakeCombination.ts
```diff
@@ -0,0 +1,80 @@
...
+ */
+
+import { VNode } from "preact";
+import { html } from "htm/preact";
+
+import { BlissSymbolInfoType, LayoutInfoType } from "./index.d";
+import { BlissSymbol } from "./BlissSymbol";
+import { composeWordContents, COMPOSE_AREA_ID } from "./GlobalData";
+import { generateGridStyle, speak } from "./GlobalUtils";
+import { isCombined, combineContent, uncombineContent, SymbolCompositionToSkip } from "./CursorActions";
```
cindyli: All these functions isCombined, combineContent, uncombineContent, SymbolCompositionToSkip are only used in this component. Why bother to define them in `CursorActions`?

src/client/CursorActions.ts
```diff
@@ -0,0 +1,197 @@
...
+import type {
+  ContentSignalDataType,
+  SymbolEncodingType,
+  SymbolCompositionType
+} from "./index.d";
+
+import { clamp } from "./GlobalUtils";
+
+// Append SymbolCompositionToSkip of symbols to skip navigating using moveCursor function to this list
+const SymbolCompositionToSkip = [233];
```
cindyli: Would be easier to understand to use `CombineMarkerId = 233;` with the code comment that it should be skipped when moving cursor.

src/client/CursorActions.ts
```diff
@@ -0,0 +1,197 @@
...
+
+/**
+ * Check if SymbolComposition is equal, as SymbolComposition can be an array or string.
+ * 
+ * @param {SymbolCompositionType} a - A SymbolComposition to be compared
+ * @param {SymbolCompositionType} b - A SymbolComposition to be compared
+ * @return {boolean} - result of the comparison
+ */
+function compositionIdEqual (a: SymbolCompositionType, b: SymbolCompositionType): boolean {
```
cindyli: As this function is only used internally, it doesn't need to be exported.

src/client/CursorActions.ts
```diff
@@ -0,0 +1,197 @@
...
+}
+
+/**
+ * Check if payload contains composition id to be skipped
+ * 
+ * @param {SymbolEncodingType} payload - An array of input symbols
+ * @param {Array<SymbolCompositionType>} skipCompositionIds - An array of composition ids to be skipped
+ * 
+ */
+function isSkipSymbol (payload: SymbolEncodingType, skipCompositionIds: Array<SymbolCompositionType>): boolean {
```
cindyli: As there's one symbol to be skipped, keep the logic simple. This function may not be needed.

src/client/CursorActions.ts
```diff
@@ -0,0 +1,197 @@
...
+  const { payloads, caretPosition } = contentSignal;
+  if (payloads.length === 0) return contentSignal;
+
+  return {
+    payloads: [combineSymbol, ...payloads, combineSymbol],
+    caretPosition: caretPosition === -1 ? 1 : caretPosition + 1
+  };
+}
+
+function uncombineContent (contentSignal: ContentSignalDataType, skipCompositionIds: Array<SymbolCompositionType>): ContentSignalDataType {
```
cindyli: Add proper code comment for functions, components and important steps to improve the code readability.

This comment applies to the entire PR.

src/client/ToggleCombineMarker.ts
```diff
@@ -0,0 +1,80 @@
...
+import { BlissSymbol } from "./BlissSymbol";
+import { composeWordContents, COMPOSE_AREA_ID } from "./GlobalData";
+import { generateGridStyle, speak } from "./GlobalUtils";
+import { isCombined, combineContent, uncombineContent, SymbolCompositionToSkip } from "./CursorActions";
+import "./ActionModifierCell.scss";
+
+const COMBINE_MARKER_PAYLOAD = {
+  "id": "foo",
+  "label": "",
+  "composition": 233,
```
cindyli, line 26: The ID for the combine marker is used in various scripts. It should be defined in the `GlobalData` to avoid the duplication.

src/client/ToggleMakeCombination.ts
```diff
@@ -0,0 +1,80 @@
...
+import "./ActionModifierCell.scss";
+
+const COMBINE_MARKER_PAYLOAD = {
+  "id": "foo",
```
cindyli: Give a meaningful id such as "combine-marker".

src/client/CursorActions.ts
```diff
@@ -0,0 +1,197 @@
...
+  if (newCaretPosition === -1 && newEncodingContents.length > 0 && isSkipSymbol(newEncodingContents[0], skipCompositionIds)) {
+    let updatedPosition = 1;
+    while (updatedPosition < newEncodingContents.length && isSkipSymbol(newEncodingContents[updatedPosition], skipCompositionIds)) {
+      updatedPosition += 1;
+    }
+
+    if (updatedPosition >= newEncodingContents.length) {
+      // Only skip symbols in the payloads, remove the symbols and reset
+      return { payloads: [], caretPosition: -1 };
+      // speak(label);
```
cindyli: Remove commented code.

src/client/CursorActions.ts
```diff
@@ -0,0 +1,197 @@
...
+function uncombineContent (contentSignal: ContentSignalDataType, skipCompositionIds: Array<SymbolCompositionType>): ContentSignalDataType {
+  const { payloads, caretPosition } = contentSignal;
+
+  const firstCombineIndex = payloads.findIndex(p => isSkipSymbol(p, skipCompositionIds));
+  let lastCombineIndex = -1;
+  for (let i = payloads.length -1; i >=0; i--) {
+    if (isSkipSymbol(payloads[i], skipCompositionIds)) {
+      lastCombineIndex = i;
+      break;
+    }
```
cindyli: Can this loop be replaced by `findLastIndex()`?

src/client/CursorActions.ts
```diff
@@ -0,0 +1,197 @@
...
+ * @return {boolean} - result of the comparison
+ */
+function compositionIdEqual (a: SymbolCompositionType, b: SymbolCompositionType): boolean {
+  if (Array.isArray(a) && Array.isArray(b)) {
+    return a.length === b.length && a.every((v, i) => v === b[i]);
+  }
+  if (Array.isArray(a) || Array.isArray(b)) {
+    return false;
+  }
+  return a == b;
```
cindyli: Use `===` instead of `==`.

src/client/CursorActions.ts
```diff
@@ -0,0 +1,197 @@
...
+  // Walk the caret left, skipping over any skip symbols
+  let newCaretPosition = caretPosition - 1;
+  while (newCaretPosition >= 0 && isSkipSymbol(newEncodingContents[newCaretPosition], skipCompositionIds)) {
+    newCaretPosition -= 1;
+  }
+
+  // If the position fell off the left side but the array still leads with a skip
+  // the caret is now outside the wrap. Try to find a symbol inside the wrap by going right
+
+  if (newCaretPosition === -1 && newEncodingContents.length > 0 && isSkipSymbol(newEncodingContents[0], skipCompositionIds)) {
```
cindyli: Line 120-133 performs the same function as `moveCursor()` function to find the next non-skip symbol. Can you extract the logic into a shared function `findNextNonSkip()`?

src/client/CursorActions.ts
```diff
@@ -0,0 +1,197 @@
...
+ * 
+ * @param {Array<SymbolEncodingType>} payloads - An array of input symbols
+ * @param {Array<SymbolCompositionType>} skipCompositionIds - An array of composition ids to be skipped
+ * 
+ */
+function isCombined(payloads: Array<SymbolEncodingType>, skipCompositionIds: Array<SymbolCompositionType>): boolean {
+  return (
+    payloads.length >=2 &&
+		isSkipSymbol(payloads[0], skipCompositionIds) &&
+		isSkipSymbol(payloads[payloads.length - 1], skipCompositionIds)
```
cindyli: These two lines use tab instead of two spaces for indentation. Please fix. Can you also check if there are other places that tabs are used? Thanks.

cindyli (changes requested):

cindyli:
Line 38 & 47 in `ActionRemoveModifierCell` component still uses the value from `changeEncodingContents`. It should be changed to use `contentSignalMap` & `isComposing` like `ActionModifierCellCommon`.

src/client/CursorActions.ts
chosww: I tried to do this but then the function needs many additional information to make it general, so the function have too many parameters:

```findNextNonSkip(payloads: Array<SymbolEncodingType>, combineSymbolId: number = combineSymbolId, direction: number, currentPosition: number, min: number, max: number): number ```

I think it's better not to extract the logic, but I if you think it's fine to have a lengthy function, I will update it.

File: src/client/CursorActions.ts (hunk 24 above)
Comment (chosww) [reply src/client/CursorActions.ts]: They are not manually tabbed, it is caused by entering a new line -- editorconfig have indent size set to 2, so all tabs have the space of 2 spaces. Lint check actually checks for incorrect spacings, and there's no error with tabs; do you want me to go ahead and fix all tabs?

File: src/client/CursorActions.ts (hunk 23 above)
Comment (cindyli) [reply src/client/CursorActions.ts]: Good point, @chosww. I think the helper function could be more focused on extracting just the while loop that finds the position of the next non-skip symbol. This logic is currently repeated at lines 77, 113, and 122.

It might be better to keep the min/max checks outside of the helper. The helper looks something like:

```
function findNextNonSkipPosition(payloads: Array<SymbolEncodingType>, startIndex: number, direction: number, skipCompositionIds: Array<SymbolCompositionType>) {
  let currentIndex = startIndex;
  while (currentIndex >= 0 && currentIndex < payloads.length && isSkipSymbol(payloads[currentIndex], skipCompositionIds)) {
    currentIndex += direction;
  }
  return currentIndex;
}
```

File: src/client/CursorActions.ts (hunk 24 above)
Comment (cindyli) [reply src/client/CursorActions.ts]: Would be nice to replace tabs with 2 spaces, at least in the new code. If there are too many tabs in the old code, we probably should address them separately.

BTW, the project has `indent_style = space` in `.editorconfig`. This should tell editors to insert spaces rather than tabs for indentations. I wonder if it's the editor setting overrides the config file. VS Code has a "insertSpace" setting that should be set to `true`.

File: src/client/CursorActions.ts (hunk 23 above)
Comment (chosww) [reply src/client/CursorActions.ts]: Thanks Cindy; though, this function works well for `deleteAtCaret` function, the while loop in move cursor has a break condition that has a different min value, as well as the while condition including the max value (<=) causes unwanted effect when moving cursors in the compose area.

File: src/client/GlobalUtils.ts (hunk 6 above)
Comment (chosww) [reply src/client/GlobalUtils.ts:139]: I think currently we only use it for modifiers, but I think Joseph had usage of this function other than adding modifier, as it does have different logic if `isModifier` is false.

File: src/client/ContentEncodingInputField.ts (hunk 8 above)
Comment (chosww) [reply src/client/ContentEncodingInputField.ts:67]: These functions take a signal as a parameter, and functions in CursorActions take plain state and returns plain state; I think mixing two signatures in one module makes it confusing, and would make the refactoring work of CursorActions meaningless, as the file is pure functions over state.

File: src/client/CursorActions.ts (hunk 23 above)
Comment (cindyli) [reply src/client/CursorActions.ts]: The check on min and max values in moveCursor() can be:

```js
  if (direction !== 0 && skipCompositionIds.length > 0) {
    // Use the helper function to find the next position
    const candidatePosition = findNextNonSkipPosition(payloads, newPosition, direction, skipCompositionIds);
    
    // If the next position pushes us out of bounds, revert to original caretPosition
    newPosition = (candidatePosition < min || candidatePosition > max) ? caretPosition: candidatePosition;
  }
```

src/client/GlobalUtils.ts
cindyli: If it's currently only used for adding modifiers and not used in any other cases, we should remove the logic when `isModifier === false` because that's probably the stale logic.

src/client/ContentEncodingInputField.ts
cindyli: Good point. I agree. Let's keep as is. Thanks.

src/client/CursorActions.ts
chosww: I tried using findLastIndex, and lint check was not happy about it -- I think it's the problem with the function itself.

src/client/CursorActions.ts
chosww: Thanks Cindy, that worked well, but I changed the function name to `findNextNonCombineSymbol` as the function is explicitly checking for combine symbols.

src/client/ActionIndicatorCell.ts
```diff
@@ -50,40 +54,46 @@ export function ActionIndicatorCell (props: ActionIndicatorCodeCellPropsType): V
...
   };
 
   return html`
-    <button id="${props.id}" class="actionIndicatorCell" style="${gridStyles}" onClick=${cellClicked} disabled="${disabled}">
+    <button
+      id="${props.id}"
+      class="actionIndicatorCell"
+      style="${gridStyles}"
+      onClick=${cellClicked}
+      disabled="${disabled}"
```
cindyli: The boolean attribute value should not be wrapped in `"`. Please remove the quotes around `${disabled}`.

Could you also take the chance to fix all such instances in the code base? Thanks.

src/client/CommandAddComposition.ts
```diff
@@ -0,0 +1,74 @@
...
+      "id": uuidv4(),
+      "label": composedLabel.trim(),
+      "composition": composedCompositionId,
+      "modifierInfo": []
+    };
+    const { payloads, caretPosition } = changeEncodingContents.value;
+    changeEncodingContents.value = insertWordAtCaret(
+      composedPayload, payloads, caretPosition
+    );
+    speak(label);
```
cindyli: Shall `composedLabel` being announced here rather than the original `label`?

src/client/ContentEncodingInputField.ts
```diff
@@ -0,0 +1,148 @@
...
+  contentSignal.value = moveCursor(Number.NEGATIVE_INFINITY, contentSignal.value, combineSymbolId);
+};
+
+export function moveCursorToEnd (contentSignal: Signal<ContentSignalDataType>) {
+  contentSignal.value = moveCursor(Number.POSITIVE_INFINITY, contentSignal.value, combineSymbolId);
+};
+
+function handleKeyDown(event: KeyboardEvent) {
+  const element = event.target as HTMLElement;
+  const contentSignal = contentSignalMap[element.id as keyof typeof contentSignalMap];
```
cindyli: This function looks up a global `contentSignalMap`. This bypasses the component's props and breaks encapsulation.

If you move this function into `ContentEncodingInputField` component, it can use component's prop `contentsSignal`.

When defining it in the component, use `const handleKeyDown = (event: KeyboardEvent) => {...}` rather than `function handleKeyDown(..)`.

src/client/ContentEncodingInputField.ts
```diff
@@ -0,0 +1,148 @@
...
+ * @param {ContentSignalDataType} symbols: Array of symbols and caret position.
+ * @return {Array<VNode>} - Array of markup for the symbols
+ */
+function generateMarkupArray (payloadArray: Array<EncodingType>, caretPos: number): Array<VNode> {
+  // NOTE:  if there are no payloads in the `payloadArray`, the map() function
+  // immediately returns an empty array.  That is, the function passed to map()
+  // will execute only if `payloadArray.length` is non-zero -- there is no need
+  // to check for a length of zero within the mapping function.
+  return payloadArray.map((payload, index) => {
+    // Check inserting before first symbol
```
cindyli, line 50: Line 51-72 can be simplified to:

```js
    const isInsertion = index === 0 && caretPos === -1;
    const isCursor = index === caretPos;
    
    // Combine classes based on conditions
    let className = "blissSymbol";
    if (isInsertion) className += " insertionCaret";
    if (isCursor) className += " cursorCaret";

    return html`
      <div class="${className}">
        <${BlissSymbol} composition=${payload.composition} label=${payload.label} isPresentation="true" />
      </div>
    `;
```

src/client/ContentEncodingInputField.ts
```diff
@@ -0,0 +1,148 @@
...
+    incrementCursor(contentSignal);
+    speak("forward");
+  }
+
+  if (
+    event.key === "Home" ||
+    (event.ctrlKey && event.key === "a") ||
+    (isApplePlatform && event.metaKey && event.key === "ArrowLeft")
+  ) {
+    event.preventDefault();
```
cindyli: `event.preventDefault()` is called for "Home" and "End" keys but not for arrow keys. If the web page has a scroller bar, hitting arrow keys will move the web page while moving the cursor. To prevent this, `event.preventDefault()` should be moved out of the `if` condition and apply for all key presses defined in this function.

src/client/ContentEncodingInputField.ts
```diff
@@ -0,0 +1,148 @@
...
+
+import { BlissSymbol } from "./BlissSymbol";
+import { contentSignalMap } from "./GlobalData";
+import { ContentEncodingType, EncodingType, ContentSignalDataType } from "./index.d";
+import { generateGridStyle, speak } from "./GlobalUtils";
+import { combineSymbolId } from "./GlobalData";
+import { moveCursor } from "./CursorActions";
+import "./ContentEncoding.scss";
+
+const isApplePlatform = navigator.platform.startsWith("Mac") || navigator.platform.startsWith("iPhone") || navigator.platform.startsWith("iPad");
```
cindyli: `navigator.platform` has deprecated. See https://developer.mozilla.org/en-US/docs/Web/API/Navigator#deprecated_properties

Replace it with `navigator.userAgentData` or other methods.

src/client/GlobalUtils.ts
```diff
@@ -132,10 +135,77 @@ function clamp (value: number, min: number, max: number) {
...
+
+  // Get the symbol at the caret position.
+  const { caretPosition, payloads } = encodingContents;
+
+
+  const symbolToEdit = payloads[caretPosition];
+  let newCompositionId = normalizedCompositionIdToAdd;
+  
+  if (caretPosition < 0) {
+    symbolToEdit.id = "foobar";
```
cindyli: `foobar` is an invalid symbol ID as well as `normalizedCompositionIdToAdd = payloads[-1]`. If the execution ever hits here, fatal errors will be reported. Currently it's not reachable because the button state is disabled when `caretPosition === -1`.

Replace this logic with a real `console.error` report rather than throwing fatal errors.

src/client/CursorActions.ts
```diff
@@ -0,0 +1,193 @@
...
+ * 
+ */
+function moveCursor (positionChange: number, contentSignal: ContentSignalDataType, combineSymbolId: number) {
+  const { payloads, caretPosition } = contentSignal;
+  const max = payloads.length - 1;
+  // If the payloads are composed symbol, the caret must stay inside the combine symbols
+  // disallow -1 as a landing position. Otherwise -1 remains valid.
+  const min = isCombined(payloads, combineSymbolId) ? 0 : -1;
+
+  let newPosition = clamp(caretPosition + positionChange, min, max);
```
cindyli, line 67: When combine marks are added at the start and end, using home/end to go to the start or end of the composing word will break.

Steps to re-produce:
Assuming `payloads=[combineMarker, A, B, combineMarker]`, `caretPosition=1`:
press End -> `positionChange=+Infinity` -> `clamp(1+Infinity,0,3)=3` -> `findNextNonCombineSymbol` walks from index 3 (the trailing marker) to index 4, which is out of bounds -> so the code reverts newPosition to the original caretPosition=1 instead of landing on 2 (B).

Handle both cases when combine markers are added and not added. Please add test cases for them. Thanks.

src/client/CommandDelLastEncoding.ts
```diff
@@ -25,26 +27,17 @@ type CommandDelLastEncodingProps = {
...
-      newEncodingContents.splice(caretPosition, 1);
-      changeEncodingContents.value = {
-        payloads: newEncodingContents,
-        caretPosition: caretPosition - 1
-      };
+    const contentSignal = contentSignalMap[ariaControls as keyof typeof contentSignalMap];
+    const updatedContentSignal = deleteAtCaret(contentSignal.value, combineSymbolId);
+    if (updatedContentSignal !== contentSignal.value) {
+      contentSignal.value = updatedContentSignal;
+      speak(label);
```
cindyli: `speak()` should be out of the loop so even when nothing to be deleted, users still get an announcement that the deletion has activated.

src/client/CommandClearEncoding.ts
```diff
@@ -25,12 +26,12 @@ type CommandClearEncodingProps = {
...
 export function CommandClearEncoding (props: CommandClearEncodingProps): VNode {
   const { id, options } = props;
-  const { label, bciAvId, columnStart, columnSpan, rowStart, rowSpan, ariaControls } = options;
+  const { label, composition, columnStart, columnSpan, rowStart, rowSpan, ariaControls } = options;
 
   const gridStyles = generateGridStyle(columnStart, columnSpan, rowStart, rowSpan);
 
   const cellClicked = (): void => {
-    changeEncodingContents.value = {
+    contentSignalMap[ariaControls as keyof typeof contentSignalMap].value = {
```
cindyli: `contentSignalMap[ariaControls as keyof typeof contentSignalMap]` is an unguarded lookup and cast. There should be a validation on `ariaControls` in case there's a missing `ariaControls` value.

This usage is also in `CommandCursorBackward`, `CommandCursorForward`, `CommandDelLastEncoding` and `ContentEncodingInputField` components.

It would be helpful to write a guarding utility function in `GlobalData.ts`:

```js
export function getContentSignal (ariaControls: string): Signal<ContentSignalDataType> | undefined {
  if (Object.prototype.hasOwnProperty.call(contentSignalMap, ariaControls)) {
    return contentSignalMap[ariaControls as keyof typeof contentSignalMap];
  }
  console.warn(`No content signal registered for ariaControls/id "${ariaControls}"`);
  return undefined;
}
```

Then change this line to:
```js
const contentSignal = getContentSignal(ariaControls);
if (!contentSignal) {
  return;
}
```

Moreover, there isn't use case that tests when `ariaControls` value is missing. Add a test for it.

src/client/Palette.integration.test.ts
```diff
@@ -35,32 +36,34 @@ describe("Palette integration test", () => {
         "type": "CommandDelLastEncoding",
         "options": {
           "label": "Delete",
-          "bciAvId": 12613,
+          "composition": 145,
```
cindyli, line 39: Although the integration test have `CommandDelLastEncoding` component added but didn't really test its clicking and interaction with other components.

Same comment applies to other components added in this PR.

src/client/ActionModifierCellCommon.ts
```diff
@@ -28,61 +29,38 @@ export type ActionModifierCodeCellPropsType = {
...
+  const modifierComposition = (
+    typeof composition === "number" ?
+      [composition] :
+      composition
   );
 
   const gridStyles = generateGridStyle(columnStart, columnSpan, rowStart, rowSpan);
-  const disabled = changeEncodingContents.value.caretPosition === -1;
+  const ariaControls = ( isComposing.value ? COMPOSE_AREA_ID : INPUT_AREA_ID);
+  const contentsSignal = contentSignalMap[ariaControls];
```
cindyli, line 44: These two lines are a copy-paste in 5 Action cell files. Write a shared helper function in `GlobalUtils.ts` for it, together with guarding against an invalid `isComposing` state.

src/client/ActionIndicatorCell.ts
```diff
@@ -27,20 +28,23 @@ export function ActionIndicatorCell (props: ActionIndicatorCodeCellPropsType): V
...
     columnStart, columnSpan, rowStart, rowSpan, label
   } = props.options;
-  const indicatorBciAvId = props.options.bciAvId;
+  const indicatorId = props.options.composition as number;
 
   const gridStyles = generateGridStyle(columnStart, columnSpan, rowStart, rowSpan);
-  const disabled = changeEncodingContents.value.caretPosition === -1;
+  const ariaControls = ( isComposing.value ? COMPOSE_AREA_ID : INPUT_AREA_ID);
+  const contentsSignal = contentSignalMap[ariaControls];
+  const disabled = contentsSignal.value.caretPosition === -1;
```
cindyli, line 36: `isComposing` and `contentSignalMap` are global signals. Every change will trigger the recalculation of `disabled` then triggers the re-render of the component.

It would be helpful to use `useComputed` to turn it into a local scoped signal:

```js
// Add at the beginning import area
import { useComputed } from "@preact/signals";

// change this line to:
const disabled = useComputed(() => contentsSignal.value.caretPosition === -1).value;
```

The re-render will only be triggered when the computed value is changed.

The same comment applies to the calculation of `disabled` in these components: `ActionModifierCellCommons`, `ActionRemoveIndicatorCell`, `ActionRemoveModifierCell`.

src/client/ActionRemoveModifierCell.ts
```diff
@@ -26,15 +27,17 @@ export function ActionRemoveModifierCell (props: ActionRemoveModifierPropsType):
...
   // input field (if any) has a modifier AND if there is more than one symbol in
   // the encoding.
   let disabled = true;
-  const { payloads, caretPosition } = changeEncodingContents.value;
+  const ariaControls =  ( isComposing.value ? COMPOSE_AREA_ID : INPUT_AREA_ID );
+  const contentsSignal = contentSignalMap[ariaControls];
+  const { payloads, caretPosition } = contentsSignal.value;
   if (payloads.length !== 0 && caretPosition !== -1) {
     const caretSymbol = payloads[caretPosition];
     disabled = !caretSymbol.modifierInfo || caretSymbol.modifierInfo.length === 0;
```
cindyli, line 43: Regarding applying `useComputed` to control `disabled` using local-scoped signal, replace these lines with:

```js
const ariaControls =  ( isComposing.value ? COMPOSE_AREA_ID : INPUT_AREA_ID );
const contentsSignal = contentSignalMap[ariaControls];
const disabled = useComputed(() => {
  const { payloads, caretPosition } = contentsSignal.value;
  if (payloads.length !== 0 && caretPosition !== -1) {
    const caretSymbol = payloads[caretPosition];
    return !caretSymbol.modifierInfo || caretSymbol.modifierInfo.length === 0;
  }
  return true;
}).value;
```

cindyli (changes requested):

cindyli:
In `index.js`, line 113 in the function `elementAllowsTextEntry()`, the check is only on `INPUT_AREA_ID`. `COMPOSE_AREA_ID` should be added so the backtick shortcut key also works when the focus is in the compose area.

src/client/ContentEncodingInputField.ts
chosww: Hey Cindy, the arrow function suggestion is because the function is used only in the component itself? if so, should I change `generateMarkupArray` to use arrow function as well?

cindyli: The suggestion of moving `handleKeyDown()` into the component definition function `ContentEncodingInputField()` is not only because it's only used in the component itself, but also because this function is reaching out to access a global variable `contentSignalMap`. For encapsulation, a function should process pass-in parameters instead of reaching out to global vars.

Moving this function into the component function `ContentEncodingInputField()` will solve this issue. `ContentEncodingInputField()` has a pass-in prop `contentsSignal` that can replace the use of `contentSignalMap`.

You don't have to change `generateMarkupArray` to use arrow function.

PR #66, opened 2026-04-24
fix: update npm packages to latest versions

cindyli:

## Description

Update all npm packages to latest versions.

The version of `@huggingface/transformers` package is still `3.8.1` while the latest version is `4.2.0` because `@langchain/community` package is dependent on this version.

## Steps to test

1. Open a terminal,
2. Make sure node and npm versions are up to date. Currently, in node 25.9.0 & npm 11.12.1
3. Run `rm -rf /node_modules`
4. Run `npm ci`
5. Run `npm outdated`. Only `@huggingface/transformers` package is reported with `3.8.1` version.
6. Run `npm run test`. All tests should pass.
7. Run `npm run lint`. All should pass.
8. Run `npm start` and test the web page. All should work well.

**Expected behavior:** 
Described with steps.

## Additional information

When running `npm start`, a warning regarding `bliss-svg-builder` package is reported. This issue is rooted to the original package and has been reported: https://github.com/hlridge/bliss-svg-builder/issues/20


.github/workflows/ci.yml
```diff
@@ -9,7 +9,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest]
-        node-version: [18.x, 20.x]
+        node-version: [24.x]
```
chosww: It would be great to update nvmrc to 24 as well.

src/client/GlobalData.ts
```diff
@@ -76,11 +77,11 @@ export const adaptivePaletteGlobals = {
   // The map between the BCI-AV IDs and the code consumed by the Bliss SVG
   // builder.  The map itself is set asynchronously.
   blissaryIdMapUrl: "https://raw.githubusercontent.com/hlridge/Bliss-Blissary-BCI-ID-Map/main/blissary_to_bci_mapping.json",
-  blissaryIdMap: null,
+  blissaryIdMap: null as BlissaryMapEntryType[] | null,
```
chosww, line 80: Can this be just `blissaryIdMap: BlissaryMapEntryType[] | null`?

src/client/GlobalData.ts
```diff
@@ -76,11 +77,11 @@ export const adaptivePaletteGlobals = {
...
   bciAvSymbols: bliss_symbols,
   paletteStore: new PaletteStore(),
   navigationStack: new NavigationStack(),
-  LLMs: [],
+  LLMs: [] as string[],
```
chosww, line 84: Can this be just `LLMs: string[]`?

chosww (commented):
Other than the server side test, everything looks good. Just a few suggestions from me.

src/client/GlobalData.ts
cindyli: If `blissaryIdMap: BlissaryMapEntryType[] | null` you suggested, and the line suggested in the comment below: `LLMs: string[]`, are in a type definition, they are fine.

In this case for an object assignment, `blissaryIdMap: null` is to assign an initial value, and the following `as ...` is a type assertion. `BlissaryMapEntryType[] | null` is not a valid initial value.

src/client/GlobalData.ts
cindyli: See the reply above.

cindyli:
@chosww All comments have been addressed, and implemented another solution to fix the server test failure on your computer. Ready for another review.

src/client/CommandGoBackCell.ts
```diff
@@ -61,7 +61,7 @@ export async function goBackImpl (defaultContaineId?: string ): Promise<void> {
 const goBackToPalette = async (event: Event): Promise<void> => {
   const button = event.currentTarget as HTMLElement;
   speak(button.innerText);
-  return goBackImpl(button.getAttribute("aria-controls"));
+  return goBackImpl(button.getAttribute("aria-controls") ?? undefined);
```
jobara: Is the nullish coalescing necessary, it's just setting the value to `undefined` if it was already null or undefined. It looks like `goBackImpl` also does nullish coalescing to set it to `""`. So it should be able to handle either `null` or `undefined` already. Or you could set it to `""` here too.

src/client/NavigationStack.ts
```diff
@@ -61,7 +61,7 @@ export class NavigationStack {
     if (this.isEmpty()) {
       return null;
     } else {
-      return this.navigateBackStack.pop();
+      return this.navigateBackStack.pop() ?? null;
```
jobara: Is it necessary to force this to be `null` if it was `undefined`?

src/client/ActionBranchToPaletteCell.ts
```diff
@@ -33,13 +33,13 @@ const navigateToPalette = async (event: Event): Promise<void> => {
   const button = event.currentTarget as HTMLElement;
   speak(button.innerText);
 
-  const buttonsPaletteName = button.parentElement.getAttribute("data-palettename");
-  const branchToPaletteName = button.getAttribute("data-branchto");
+  const buttonsPaletteName = button.parentElement!.getAttribute("data-palettename")!;
+  const branchToPaletteName = button.getAttribute("data-branchto")!;
```
jobara: If I'm understanding correctly the `!` used here are for typescript to avoid type checking around null and undefined. Saying that these aren't those values. Is that correct, and is that safe to assume for runtime?

src/client/CommandGoBackCell.ts
cindyli: You’re right from a runtime perspective. However, this change is to address a TypeScript [`strictNullChecks`]((https://www.typescriptlang.org/tsconfig/#strictNullChecks)) error at compile time.

With strictNullChecks enabled, TypeScript considers `null` and `undefined` as two different types. The issue here is `button.getAttribute("aria-controls")` returns `string | null`, while `goBackImpl()` expects a parameter of type `string | undefined`. We can’t directly pass the result of `null` into goBackImpl().

I think a cleaner fix is to update the `goBackImpl()` to accept `null` that is more align with how DOM works because DOM likes to return `null`. This helps to remove the nullish coalescing when preparing the parameter for `goBackImpl()`.

src/client/NavigationStack.ts
cindyli: See the comment above. It is necessary because typescript treats `null` and `undefined` as two different types.

src/client/ActionBranchToPaletteCell.ts
cindyli: You are right: it's not safe for runtime. I've fixed it.

cindyli:
@jobara @chosww all review comments have been addressed. Ready for more reviews.

src/client/ActionBranchToPaletteCell.ts
```diff
@@ -33,13 +33,17 @@ const navigateToPalette = async (event: Event): Promise<void> => {
   const button = event.currentTarget as HTMLElement;
   speak(button.innerText);
 
-  const buttonsPaletteName = button.parentElement.getAttribute("data-palettename");
+  const buttonsPaletteName = button.parentElement!.getAttribute("data-palettename")!;
```
jobara: It would probably be better to handle all of the cases where `!` is used to ensure that either they aren't null/undefined, or if they are that they are handled appropriately.

src/client/GlobalUtils.ts
```diff
@@ -22,7 +22,7 @@ import { JsonPaletteType, SymbolEncodingType, ContentSignalDataType } from "./in
  * @param {number} rowSpan - The number of rows that the item will span across.
  * @return {String} - The grid css.
  */
-function generateGridStyle(columnStart: number, columnSpan: number, rowStart: number, rowSpan: number): string {
+function generateGridStyle(columnStart: number | undefined, columnSpan: number | undefined, rowStart: number | undefined, rowSpan: number | undefined): string {
```
jobara: What happens if these values are undefined? Would it interpolate the string "undefined" into the output or have otherwise invalid CSS? If that's the case you should either use a default value for those parameters or have logic to only interpolate actual values into the output.

src/client/NavigationStack.ts
jobara: I think part of the problem here is that similar functions like peek can return `undefined` instead of `null`. These things that tend to return `NavStackItemType` should probably all have the same return types.

src/client/NavigationStack.ts
```diff
@@ -61,7 +61,7 @@ export class NavigationStack {
...
     }
   }
```
jobara, line 63: It isn't necessary to have an if/else here. Here are a couple of potential simplifications:

Just remove the else, as it will fall down to the default return anyways.

```javascript
  pop (): NavStackItemType | null {
    if (this.isEmpty()) {
      return null;
    } 
    return this.navigateBackStack.pop() ?? null;
  }
```

Remove the conditional entirely, as `pop()` on an array returns undefined if it's empty.

```javascript
  pop (): NavStackItemType | null {
    return this.navigateBackStack.pop() ?? null;
  }
```

If you change the return of `pop` to allow for `undefined` instead of `null` (see other comment about consistent return values), you can just return the underlying call directly.

```javascript
  pop (): NavStackItemType | undefined {
    return this.navigateBackStack.pop();
  }
```

jobara (changes requested):

cindyli:
@jobara all new review comments have been addressed. Ready for another look. Thanks.

PR #89, opened 2026-05-06
Feat: enable type check (resolves #45 & #59)

### LINKED ISSUE #45: Bug: The script `scripts/createAndRecordCompositions.ts` no longer functions
### Describe the bug

`scripts/createAndRecordCompositions.ts` is a one-time script for creating the initial `public/data/bliss_symbol_explanations.json`. It imports a function `makeBlissComposition()` that has been removed from the code base so this script no longer functions.

`bliss_symbol_explanations.json` will be replaced by a multi-language supported file. Besides that, if there will be changes to this file in the future, most likely they will be minor adjustments and we'll do it manually. 

Since the script is no longer needed and already broken, we should consider removing it from the code base to reduce maintenance work and avoid confusion.

### To reproduce

Check the script content for the missing import function.

### Expected behaviour

Evaluate if this script is needed. If not, remove it. If yes, fix it.

### Screenshots

_No response_

### Desktop

_No response_

### Mobile

_No response_

### Additional context

_No response_

### LINKED ISSUE #59: Feature: Enable the type check

VS Code reports type check errors, while `npm run lint` doesn't although the typescript check in the eslint has been enabled.

This is because:

1. VS Code has a built-in TypeScript Language Server that continuously runs the equivalent of the TypeScript compiler (tsc) in the background when someone is typing. It checks for type validity such as "You are passing a string to a function that expects an number".
2. ESLint is a linter. Even if having `@typescript-eslint` installed, ESLint primarily looks at the syntax of the code to find stylistic issues, bad practices, and potential bugs such as "You have unused variables" or "You are missing a semicolon". By default, ESLint does not check TypeScript types. It assumes the code is type-safe and only checks it against linting rules.

1. Add a `typecheck` job into `package.json` that runs `tsc --noEmit` to check type errors.
2. Enable "Typed Linting" in ESLint by including `plugin:@typescript-eslint/recommended-requiring-type-checking'`. This cannot replace `tsc` but it unlock ESLint rules like warning you if you forget to await a Promise.
3. For those who use VS Code, we also need to sync the VS Code version with this project. Steps:
    (1) Open any .ts file in VS Code.
    (2) Open the Command Palette (Ctrl+Shift+P on Windows/Linux, Cmd+Shift+P on Mac).
    (3) Search for and select: TypeScript: Select TypeScript Version.
    (4) Choose Use Workspace Version.

cindyli:
# Pull Request Checklist

## Description

This pull request adds a npm job "typecheck" for type checking. Type checking checks types at compile-time and catch errors early.

## Steps to test

1. Run `npm run lint`, `npm run typecheck`, `npm run test` to make sure they all pass.
2. Run `npm start` or `npm run dev` to start the application.
3. Test the app to make sure it works properly.

**Expected behavior:** See above.


apps/palette-generator/paletteJsonGenerator.ts
```diff
@@ -108,15 +135,17 @@ function findBciAvId(label, blissGlosses) {
...
       // ignore it.
       const glossId = parseInt(gloss.id);
-      let fullComposition;
+      let fullComposition: BciAvIdType | undefined;
       let equalCompositions = false;
       if (gloss.composition) {
         fullComposition = decomposeBciAvId(gloss.composition);
-        equalCompositions = fullComposition.join("") === gloss.composition.join("");
+        const compArr = Array.isArray(gloss.composition) ? gloss.composition : [gloss.composition];
+        const fullArr = Array.isArray(fullComposition) ? fullComposition : (fullComposition !== undefined ? [fullComposition] : []);
```
jobara: `null` and `undefined` values are converted to `""` by the `join()`, so it probably isn't necessary to check for `undefined` here.

see: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/join#description

jobara: If you actually do care about `null` and/or `undefined` values, you would need a different way to compare. For example using `JSON.stringify()`; however that converts `undefined` to `null`. Or you could do a comparison of the values at each index.

apps/palette-generator/paletteJsonGenerator.ts
```diff
@@ -244,26 +279,28 @@ export function processPaletteLabels (paletteLabels, paletteName, startRow, star
...
+            const inputMatches: MatchByInfo = {};
             inputMatches[infoString] = matches;
             matchByInfoArray.push(inputMatches);
           }
         }
       }
       catch (error) {
         // If an error occurs, add it to the errors array
-        errors.push(error.message);
+        const message = error instanceof Error ? error.message : String(error);
```
jobara, line 302: Is there something that throws something other than an Error? Is it required to handle this case because something could have thrown something other than an Error even if we don't have that in the code?

apps/palette-generator/paletteJsonGenerator.ts
cindyli: `fullComposition` value is the result of `decomposeBciAvId()` that can return `undefined`, so the handling of `undefined` is required here.

apps/palette-generator/paletteJsonGenerator.ts
cindyli: Javascript has no type enhancement for `error`. It can literally be anything: a string, an object or other data types. `instanceof Error` is to tell typescript the data type of `error` and it has an attribute of `message`. Otherwise, typescript will consider `error` and `error.message` use type `any` which will cause a linting error of `@typescript-eslint/no-unsafe-argument` because by default typescript prevents the use of `any` type.

cindyli:
@jobara please see my replies to your comments and let me know what you think. Thanks.

apps/palette-generator/paletteJsonGenerator.ts
jobara: Hi @cindyli what I mean is that when `fullArr.join("")` is called both `undefined` and `null` values will both be collapsed to "". 

for example:

```JavaScript
let arr = [1, "a", undefined, null];
arr.join(""); // result is "1a"
```

apps/palette-generator/paletteJsonGenerator.ts
jobara: If you look in the codebase there are a lot of try/catch blocks, but this is the only case where the error instance is checked because it's also the only case that the message is accessed off of it.

There are a variety of ways that the catch is handled: `console.log`, `console.error`, `console.debug` for example and some include extra text, others just the error. However, this is the only case that tries to access the message off of the error object. Unless this is a special case we should either also check the other places that are used, at least when interpolated into a string, or in this case just pass in the object too. It would be worth ensuring that when interpolated into a string that the output doesn't say `[object object]` or something like that. 

see: https://github.com/search?q=repo%3Ainclusive-design%2Fadaptive-palette%20catch&type=code

cindyli: This line is a special case because `errors` is `string[]` so the string extraction is needed here.

I checked other places. One issue is with `server.ts` at line 56-58 where `Error` is an object. Just fixed it by adding string extraction.

apps/palette-generator/paletteJsonGenerator.ts
cindyli: You are right. The check of `undefined` has been removed.

cindyli:
@jobara your comments have been addressed. Ready for another look. Thanks.

PR #112, opened 2026-05-19
Chore: Update `bliss_symbol_explanations.json` to latest Bliss symbol set (resolves #111)

### LINKED ISSUE #111: Chore: Update `bliss_symbol_explanations.json` to latest Bliss symbol set
## Problem
The file `public/data/bliss_symbol_explanations.json` contains Bliss symbol metadata, including BCI-AV-ID, gloss, part of speech (pos), explanation.

It is currently outdated and does not reflect the latest Bliss symbol set.

## Solution
Hannes has an updated version of the Bliss symbol set (as of 2026-05-13), which includes all data required by the symbol explanation file. We should update `bliss_symbol_explanations.json` using Hannes’ latest dataset.

cindyli:
# Pull Request Checklist

## Description

Update `public/data/bliss_symbol_explanations.json` to the latest Bliss symbol set. It also takes the chance to rename the previous "description" attribute name in the file to "gloss".

The update is based on Hannes' latest data set located at `scripts/data/bliss_dictionary-20260513.json`.

## Steps to test

1. Run regular test steps: `npm run lint`, `npm run typecheck`, `npm run test`;
2. Run `npm start` to start the app at `http://localhost:3000/`

**Expected behavior:** Test everything to make sure they work well.


apps/palette-generator/paletteJsonGenerator.ts
```diff
@@ -116,30 +106,30 @@ function convertSvgBuilderString (theString: string): BciAvIdType {
...
  */
-function findBciAvId(label: string, blissGlosses: BlissGlossEntry[]): MatchInfo[] {
+function findBciAvId(label: string, blissSymbolEntries: BlissSymbolEntry[]): MatchInfo[] {
   const matches: MatchInfo[] = [];
   // Search for the label in the Bliss gloss
   console.log(`For label ${label}:`);
-  for (const gloss of blissGlosses) {
+  for (const oneBlissSymbolEntry of blissSymbolEntries) {
     // Try an exact match or a word match
     const wordMatch = new RegExp("\\b" + `${label}` + "\\b");
```
jobara, line 122: This is a bit off topic but noticing the two forms of string concatenation here, and wondering if it could be written like:

```suggestion
    const wordMatch = new RegExp(`\\b${label}\\b`);
```

scripts/generate_bliss_symbol_explanations.js
```diff
@@ -0,0 +1,470 @@
...
+ * Usage:
+ * node generate_bliss_symbol_explanations.js <inputFile.json> <outputFile.json>
+ *
+ * Example:
+ * node generate_bliss_symbol_explanations.js data/bliss_dictionary_20260513.json ../public/data/bliss_symbol_explanations.json
+ *
+ * This script processes a JSON file containing linguistic derivation data and 
+ * maps it into a new hierarchical structure used by this project. The original
+ * data file is used by bliss-svg-builder: https://github.com/hlridge/bliss-svg-builder
+ * and located at scripts/data/bliss_dictionary-20260513.json.
```
jobara, line 11: Is this file included in an npm package, either on it's own or part of another package. If so, maybe we can could just pull it in from there instead of including it directly.

scripts/generate_bliss_symbol_explanations.js
```diff
@@ -0,0 +1,470 @@
...
+ *      - `code`: Decomposes the code into its constituent parts.
+ *    1.2 Fallback for missing `isChar`: defaults to `false`.
+ * 2. Composition Generation: Recursively decomposes non-character items 
+ *    (`isChar === false`) down to their base characters (`isChar === true`).
+ * 3. Delimiters: Inserts ";" before IDs that fall into the provided indicator 
+ *    ranges/lists, and "/" otherwise, dynamically flattening the array.
+ * 
+ * Manual operations after script execution:
+ * 1. Add "pos" value "noun" to item with ID "B6437"
+ * 2. Add "pos" value "expression" to item with ID "B6438"
```
jobara, line 27: Is it possible to update the original file with these changes too instead of doing the manual cleanup after?

scripts/generate_bliss_symbol_explanations.js
hlridge: It probably should since licenses are not compatible?

cindyli: @hlridge do you mean the original file from you should be removed from this pull request? Thanks.

scripts/generate_bliss_symbol_explanations.js
cindyli: @jobara The interesting for these two items is they are not at the top level definition where the "pos" value is presented. @hlridge any thoughts why they are missing at the top level? Thanks.

scripts/generate_bliss_symbol_explanations.js
hlridge: As I understand it, if you include CC-BY-SA 4.0 in a project that is BSD 3-Clause, the project becomes dual-licensed. That is fine, but for downstream users of this project, it should probably be stated in the LICENSE file in the root.

hlridge: It's a big JSON file and keeping this up-to-date in a separate GitHub repo is probably not optimal.

scripts/generate_bliss_symbol_explanations.js
hlridge: @cindyli  I forgot to add them when I added the entries to Blissary some time ago. It's been updated there, when you brought it to my attention, thanks!

scripts/generate_bliss_symbol_explanations.js
```diff
@@ -0,0 +1,470 @@
...
+  let outId;
+  if (item.bciAvId !== undefined && item.bciAvId !== null) {
+    outId = item.bciAvId.toString();
+  } else {
+    outId = "B" + item.id;
+    reports.missingTopBciAvId.add(item.id);
+  }
+
+  /** @type {{ id: string, gloss: string, pos?: string, explanation?: string, isCharacter: boolean, composition?: (string | number)[] }} */
+  const outItem = {
```
hlridge, line 421: If the TypeScript type requires these values, I think it would be good to normalize the generated output by applying defaults here.
External dictionary data will not always be 100% complete.

For example, `B6437` and `B6438` were missing `pos`, which would not be a problem if the generator used:

```js
  pos: item.pos ?? "",
```
Similarly, B6330 appears to lack explanation in the BCI-AV source data, which could be handled with:

```js
  explanation: item.explanation ?? "",
```

That way the generated JSON consistently matches the app-facing type, even when the source data has gaps.

apps/palette-generator/paletteJsonGenerator.ts
```diff
@@ -116,30 +106,30 @@ function convertSvgBuilderString (theString: string): BciAvIdType {
...
-    if ((label === gloss.description) || wordMatch.test(gloss.description)) {
+    if ((label === oneBlissSymbolEntry.gloss) || wordMatch.test(oneBlissSymbolEntry.gloss)) {
       // Get the composition of all the parts of the symbol's compostion or
       // its ID.  But if the `fullComposition` is the same as the original
       // ignore it.
-      const glossId = parseInt(gloss.id);
+      const glossId = parseInt(oneBlissSymbolEntry.id);
```
hlridge, line 127: `oneBlissSymbolEntry.id` can now contain a fallback B-code such as `"B6437"` when there is no BCI-AV ID. In that case `parseInt(oneBlissSymbolEntry.id)` returns `NaN`, so `glossId` can become `NaN`. Same concern may also apply in `apps/palette-generator/paletteJsonGenerator.ts`, where `oneBlissSymbolEntry.id` is also parsed with `parseInt`.

scripts/generate_bliss_symbol_explanations.js
```diff
@@ -0,0 +1,470 @@
...
+ * 2. Composition Generation: Recursively decomposes non-character items 
+ *    (`isChar === false`) down to their base characters (`isChar === true`).
+ * 3. Delimiters: Inserts ";" before IDs that fall into the provided indicator 
+ *    ranges/lists, and "/" otherwise, dynamically flattening the array.
+ * 
+ * Manual operations after script execution:
+ * 1. Add "pos" value "noun" to item with ID "B6437"
+ * 2. Add "pos" value "expression" to item with ID "B6438"
+ * 
+ * Execution report:
```
hlridge, line 29: Was this execution report meant to be kept in the script header? It looks like output from a particular run, so the counts and warnings may become stale when the source dictionary changes.

scripts/generate_bliss_symbol_explanations.js
hlridge: For small source-data issues like these missing `pos` values, I think it is better for the code to handle imperfect dictionary data gracefully rather than patching individual entries directly in this codebase.

Dictionary data tends to contain many small imperfections and incremental changes, so the software should not depend on every external entry being complete. A routine for updating from newer Blissary JSON snapshots seems more maintainable than one-off source-data fixes here.

hlridge (commented):
A few comments about normalizing incomplete source data, handling fallback B-code IDs, the script header containing run-specific reports, and avoiding one-off source-data fixes in the codebase.

I am also skeptical about including the source JSON data directly in this repo because of the dual-licensing complexities, as discussed in the other comments.

scripts/generate_bliss_symbol_explanations.js
cindyli: This file is not included in a npm package. It's a OneDrive file @hlridge shared. Since this file is too big to fetch in script, I'll add a download step to manually download it before running the conversion script.

cindyli:
Closed as changes in this PR have been incorporated into https://github.com/inclusive-design/adaptive-palette/pull/126. All comments have been addressed there as well.

PR #126, opened 2026-05-29
feat: update bliss_symbol_explanations.json and implement the new ID system

cindyli:
# Pull Request Checklist

## Description

* Updated `public/data/bliss_symbol_explanations.json` to align with the latest Bliss symbol set. As part of this update, the `description` field has been renamed to `gloss`.
* In the updated symbol set, 9 Bliss symbols are missing from the official BCI-AV so they do not have corresponding `bciAvId` values (previously used as keys in `public/data/bliss_symbol_explanations.json`). To address this issue and potential future same issues, a new `id` field has been introduced in the element data structure. This `id` is aligned with the Blissary ID used in BlissSvgBuilder, while `bciAvId` is now retained as a separate attribute within each element when available.
* Updated related scripts to support the new ID system.

## Steps to test

1. Run regular test steps: `npm run lint`, `npm run typecheck`, `npm run test`;
2. Run `npm start` to start the app at `http://localhost:3000/`

**Expected behavior:** Test everything to make sure they work well.


src/client/SvgUtils.ts
```diff
@@ -10,390 +10,246 @@
...
+    builder = new BlissSVGBuilder(compositionToBstr(id));
   }
   catch (err) {
     console.error(err);
-    console.error(`Unknown bci-av-id = ${String(bciAvId)}`);
+    console.error(`Unknown id = ${String(id)}`);
     builder = null;
   }
   return builder;
 }
```
hlridge, line 262: **High — `getSvgBuilder` doesn't resolve composite ids inside arrays.** A composite id nested in an array reaches the builder as a bare `B<id>` and renders empty — currently blanking `when [2734]`, `how [1832]`, `massage [4749]`, `do, does [1558]`, and dropping the base glyph from `don't know [1903,"/",449,"/",401]`. Resolve composites in both branches (one level suffices — the dataset is fully decomposed):

```suggestion
function getSvgBuilder (id: SymbolCompositionType): BlissSVGBuilder | null {
  // Composite IDs (isCharacter: false) aren't in the builder's internal
  // database and render as empty SVGs, so replace any composite ID — whether
  // passed on its own or nested in an array — with its base-character
  // composition before building.
  const resolve = (single: number): SymbolCompositionType =>
    adaptivePaletteGlobals.symbols.find(s => s.id === single)?.composition ?? single;

  if (typeof id === "number") {
    if (!adaptivePaletteGlobals.symbols.find(s => s.id === id)) return null;
    id = resolve(id);
  }
  else {
    id = id.flatMap(item => (typeof item === "number" ? resolve(item) : item));
  }
  let builder;
  try {
    builder = new BlissSVGBuilder(compositionToBstr(id));
  }
  catch (err) {
    console.error(err);
    console.error(`Unknown id = ${String(id)}`);
    builder = null;
  }
  return builder;
}
```

_Verified locally: typecheck + SvgUtils tests pass, and `[1558]`/`[2734]`/`[1832]` then render non-empty and match their scalar form. Longer term these could be registered as builder aliases instead, but that's better left to the svg-builder refactor._

apps/palette-generator/paletteJsonGenerator.ts
```diff
@@ -44,12 +32,12 @@ type ProcessPaletteResult = {
...
 
-let bliss_gloss: BlissGlossEntry[];
-export async function fetchBlissGlossJson (): Promise<BlissGlossEntry[]> {
+let bliss_gloss: BlissSymbolEntry[];
+export async function fetchBlissGlossJson (): Promise<BlissSymbolEntry[]> {
   // Read and parse the Bliss gloss JSON file
   try {
-    const fetchResponse = await fetch("/data/bliss_symbol_explanations.json");
-    bliss_gloss = await fetchResponse.json() as BlissGlossEntry[];
+    const fetchResponse = await fetch("../../public/data/bliss_symbol_explanations.json");
```
hlridge: **High — palette-generator data fetch 404s.** Vite (and the express server) serve `public/` at the URL root, so from `/apps/palette-generator/` this resolves to `/public/data/…`, which 404s — `bliss_gloss` stays undefined and every gloss/BCI-AV lookup throws into the NOT-FOUND branch. Revert to the root-served path:

```suggestion
    const fetchResponse = await fetch("/data/bliss_symbol_explanations.json");
```

scripts/generate_bliss_symbol_explanations.js
```diff
@@ -0,0 +1,227 @@
...
+ * Usage:
+ * node generate_bliss_symbol_explanations.js <inputFile.json> <outputFile.json>
+ *
+ * Example:
+ * node generate_bliss_symbol_explanations.js data/bliss_dictionary_20260513.json ../public/data/bliss_symbol_explanations.json
+ *
+ * This script processes a JSON file containing linguistic derivation data and 
+ * maps it into a new hierarchical structure used by this project. The original
+ * data file is used by bliss-svg-builder: https://github.com/hlridge/bliss-svg-builder
+ * and located at scripts/data/bliss_dictionary-20260513.json.
```
hlridge: **Medium — the generator's input isn't committed or documented, so the dataset can't be regenerated.** The input `scripts/data/bliss_dictionary_*.json` was added earlier in this PR and removed in the final commit, and there's no recorded source/URL/checksum or fetch step. The shipped data is correct so this isn't a blocker, but please either commit the input under `scripts/data/` or document its provenance (source repo + commit/date + checksum) and add a fetch step. (Minor: the header names the file inconsistently — line 6 uses an underscore, line 11 a hyphen — and neither path currently exists.)

```suggestion
 * and located at scripts/data/bliss_dictionary_20260513.json.
```

src/client/index.d.ts
```diff
@@ -9,17 +9,24 @@
...
+// Symbol composition can be either a symbol ID defined in bliss_symbol_explanations.json,
+// for example 1433, or an array of symbol IDs and/or strings that represent the composition
+// of a symbol, for example [1433, "/", 1234].
+export type SymbolCompositionType = number | (string|number)[];
 
-export type BlissaryMapEntryType = {
-  blissaryId: number,
+export type BlissSymbolEntry = {
+  id: number,
   bciAvId: number,
```
hlridge: **Medium — make `bciAvId` optional.** 9 symbols legitimately have no `bciAvId`, but this type declares it required and `GlobalData` casts the JSON with `as BlissSymbolEntry[]`, hiding the mismatch — so `.bciAvId` is `undefined` at runtime for those entries. Gloss-searching one then renders the literal text `undefined`. _(Verified: typecheck still passes on client and server with this change.)_

```suggestion
  bciAvId?: number,
```

src/client/index.d.ts
```diff
@@ -86,8 +93,8 @@ export type ContentSignalDataType = {
  * Match structure when searching the BCI AV for symbols
  */
 export type MatchType = {
+  id: number,
   bciAvId: number,
```
hlridge: Same change here — `findSymbolByGloss` copies `bciAvId` into every `MatchType`, so for the 9 keyless symbols this field is `undefined` despite the required `number` type.

```suggestion
  bciAvId?: number,
```

src/client/ActionGlossSearchCell.ts
```diff
@@ -16,61 +16,52 @@ import { BlissSymbolInfoType, LayoutInfoType } from "./index.d";
...
 import { changeEncodingContents } from "./GlobalData";
 import { generateGridStyle, speak, insertWordAtCaret } from "./GlobalUtils";
-import { decomposeBciAvId } from "./SvgUtils";
+import { compositionToBstr } from "./SvgUtils";
 import "./ActionGlossSearchCell.scss";
 
 type ActionGlossSearchCellPropsType = {
   id: string,
-  options: BlissSymbolInfoType & LayoutInfoType
+  options: BlissSymbolInfoType & LayoutInfoType & { id: number, bciAvId: number }
```
hlridge: Match the optional type on the component's props so the data contract lines up.

```suggestion
  options: BlissSymbolInfoType & LayoutInfoType & { id: number, bciAvId?: number }
```

src/client/ActionGlossSearchCell.ts
```diff
@@ -80,19 +71,19 @@ export function ActionGlossSearchCell (props: ActionGlossSearchCellPropsType): V
...
-          ref=${inputRef} 
-          id="input-${props.id}" 
-          defaultValue=${proposedGloss} 
+        <input
+          ref=${inputRef}
+          id="input-${props.id}"
+          defaultValue=${proposedGloss}
         />
-        <span>${bciAvId}: ${compositionString}</span>
+        <span>${bciAvId} . ${compositionToBstr(id)} . ${compositionString}</span>
```
hlridge: Guard the render so a keyless symbol doesn't display the literal `undefined`:

```suggestion
        <span>${bciAvId ?? ""} . ${compositionToBstr(id)} . ${compositionString}</span>
```

scripts/generate_bliss_symbol_explanations.js
```diff
@@ -0,0 +1,227 @@
...
+if (reports.missingIsChar.size > 0) {
+  console.log(`\nWarning: Items missing "isChar" key (defaulted to false): ${reports.missingIsChar.size} items.`);
+}
+
+if (reports.missingBPrefix.size > 0) {
+  console.log(`\n=== Special Code Segment Report (Total: ${reports.missingBPrefix.size}) ===`);
+  reports.missingBPrefix.forEach(msg => console.log(msg));
+}
+
+if (reports.missingBPrefix.size === 0 && reports.circularDependency.size === 0) {
```
hlridge: **Low — "No structural errors detected" can print while a Missing-ID-Reference report also prints.** The all-clear gate omits `missingIDReference`, so a future input with a dangling `B<id>` reference would print both. One-token fix:

```suggestion
if (reports.missingBPrefix.size === 0 && reports.circularDependency.size === 0 && reports.missingIDReference.size === 0) {
```

src/client/BciAvUtils.ts
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright 2025 - 2026 Inclusive Design Research Centre, OCAD University
+ * Copyright 2025-2026 - 2026 Inclusive Design Research Centre, OCAD University
```
hlridge: **Nit — mangled copyright year introduced by this PR** (`2025 - 2026` became `2025-2026 - 2026`):

```suggestion
 * Copyright 2025-2026 Inclusive Design Research Centre, OCAD University
```

hlridge (changes requested):
## Summary

Solid and well-motivated PR. It re-keys the symbol dataset from BCI-AV id to **Blissary id** (so the 9 symbols with no BCI-AV id can finally be represented), splits `bciAvId` out as a separate optional field, renames the long-mislabeled `description` field to `gloss`, and meaningfully simplifies `SvgUtils`.

I checked it out locally and the important things hold up:
- **typecheck + eslint + 307/307 tests pass**;
- the data file matches every claim — 6,420 entries, exactly 9 without a `bciAvId`, no duplicate ids, and all 5,215 non-characters carry a fully-decomposed `composition`;
- the part most likely to break silently — the **indicator/modifier id remap** — is faithful: every rewritten id resolves to the same BCI-AV value it replaced, so detection behaviour is unchanged from `main`.

Two things to fix before merge (inline):
1. **`getSvgBuilder` doesn't resolve composite ids inside arrays**, so 5 palette cells render blank or partial — `when`, `how`, `massage`, `do, does`, and `don't know` in `top_palette`. Confirmed by actually rendering them.
2. **The palette-generator's data fetch** now points at `../../public/data/…`, which 404s under Vite/express and breaks its gloss/ID lookups.

Then three mediums (optional `bciAvId` typing; commit/document the generator input so the dataset is reproducible; include `missingIDReference` in the generator's all-clear gate) and a few minor items, all inline.

Nice wins worth calling out: dropping the bidirectional Blissary↔BCI-AV map in favour of ids that line up with the builder directly, and the `ollama.js` `Promise.all` fix.

---

**Minor / non-blocking (all introduced by this PR):**
- **`findSymbolByBciAvId` name collision** — now exported from both `BciAvUtils.ts` (returns matches) and `SvgUtils.ts` (returns one entry). Worth renaming one, e.g. the composition search → `findSymbolsUsingBciAvId`.
- **`SvgUtils.findSymbolByBciAvId`** — tighten the param to `number` and skip `undefined` so it can't match a keyless symbol.
- **Generator** — the circular-dependency `continue` can orphan a separator in `composition` for cyclic input; and `!item.bciAvId` (line 153) vs `?? undefined` (line 168) disagree on `0` — use `== null` in both.
- **Tests** — add coverage for a `bciAvId`-less symbol, and for `bstrToComposition`'s `AK:` kerning / bare-digit rejection.

cindyli:
@hlridge thanks for the review. All comments have been addressed. Ready for another look.

hlridge:
> @hlridge thanks for the review. All comments have been addressed. Ready for another look.

I pulled the branch and typecheck, lint, and the full test suite (310 tests across all three browsers) pass. Everything from the last round is handled cleanly, and the getResolvedComposition extraction plus the new docs and tests are nice additions. Data migration checks out too.

Three small follow-ups, neither blocking:

1. apps/palette-generator/paletteJsonGenerator.ts: the local MatchInfo type still has `bciAvId: number`, but since BlissSymbolEntry.bciAvId is now optional, line 122 assigns `number | undefined`. tsc flags it (TS2322 at 122), but CI misses it
 because apps/ isn't in either typecheck project. Same fix as in index.d.ts: `bciAvId?: number`. Otherwise the 9 entries without a bciAvId render "undefined: …" in the output.

2. README line 220: the new "Generate …" link is dead. The path is wrapped in backticks so it isn't a link, and the filename is singular (generate_bliss_symbol_explanation.js) where the file is plural (…explanations.js).

3. In getResolvedComposition, an unknown id inside an array falls through as a raw number rather than returning null like the scalar path, so it'd render a blank glyph instead of failing, not reachable from the current data, just a small asymmetry.

cindyli:
@hlridge new comments have been addressed. Ready for another round. Thanks.

hlridge:
> @hlridge new comments have been addressed. Ready for another round. Thanks.

looks good now! just a final s missing in the filename generate_bliss_symbol_explanation[HERE].js

cindyli:
> looks good now! just a final s missing in the filename generate_bliss_symbol_explanation[HERE].js

@hlridge the "s" is already included in the script name. I also checked the code base to make sure there isn't a missing "s" in this filename. Is there somewhere else I missed? Thanks.

hlridge:
@cindyli it's at [line 220 in README.md](https://github.com/cindyli/adaptive-palette/blob/feat/implement-new-symbol-ids/README.md?plain=1#L220).

cindyli:
> @cindyli it's in https://github.com/cindyli/adaptive-palette/tree/feat/implement-new-symbol-ids under Utility scripts. the link to the json file

Good catch, @hlridge. Fixed. Thanks!

apps/palette-generator/index.html
```diff
@@ -89,12 +89,14 @@ <h1>Palette Generator</h1>
          </ul>
       </li>
       <li>
-        A single word or words separated by underscores.  The glossary is
-        searched for words that match, or for a partial or exact match with the
-        gloss of an individual Bliss symbol.  Mutitple matches are possible,
-        but only the first match is displayed in the preview.
+        A gloss.  Use
+        <code>GLOSS:<span style="font-style: italic;">gloss</span>:GLOSS</code>.
```
jobara: The styling should be moved out to the style block, and ideally that would be in a separate CSS file if possible.

jobara: You could also use a `<em></em>` tag.

jobara: Similar comment for other parts that have inline styling.

apps/palette-generator/index.html
jobara, line 1: For a future update, I think the UI for this tool could be improved. It's hard to read the syntax and understand how a grid is established from the text. Similar to how it can be hard to read table structures in markdown.

apps/palette-generator/paletteJsonGenerator.ts
```diff
@@ -78,34 +66,32 @@ function isSvgBuilderString (theString: string): boolean {
...
+  return symbol.id;
+}
+
+function convertSvgBuilderString (theString: string): SymbolCompositionType {
+  let result: SymbolCompositionType;
+  // Two forms:
+  // - no commas, using Blissary IDs with "B" prefix (e.g. "B220;B99"):
+  //   Remove the SVG prefix/suffix and use bstrToComposition().
+  // - no commas, using BCI AV IDs as plain numbers (e.g. "13166;9011"):
+  //   Parse tokens, look up each number as a BCI AV ID, return blissary IDs.
```
jobara: Is the "no commas" important still, as the "commas" portion was removed from the comment.

apps/palette-generator/paletteJsonGenerator.ts
```diff
@@ -116,46 +102,28 @@ function convertSvgBuilderString (theString: string): BciAvIdType {
...
-      }
+    const wordMatch = new RegExp(`\\b${label}\\b`);
+    if ((label === oneBlissSymbolEntry.gloss) || wordMatch.test(oneBlissSymbolEntry.gloss)) {
       matches.push({
-        bciAvId: glossId,
-        label: gloss.description,
-        composition: gloss.composition,
-        fullComposition: ( equalCompositions ? undefined : fullComposition )
+        id: oneBlissSymbolEntry.id,
+        bciAvId: oneBlissSymbolEntry.bciAvId ?? undefined,
```
jobara, line 122: Is it necessary to use nullish coalescence here? Is there an issue if the bciAvID is null instead of undefined?

apps/palette-generator/paletteJsonGenerator.ts
```diff
@@ -78,34 +66,32 @@ function isSvgBuilderString (theString: string): boolean {
...
+}
+
+function convertSvgBuilderString (theString: string): SymbolCompositionType {
+  let result: SymbolCompositionType;
+  // Two forms:
+  // - no commas, using Blissary IDs with "B" prefix (e.g. "B220;B99"):
+  //   Remove the SVG prefix/suffix and use bstrToComposition().
+  // - no commas, using BCI AV IDs as plain numbers (e.g. "13166;9011"):
+  //   Parse tokens, look up each number as a BCI AV ID, return blissary IDs.
+  if (theString.indexOf("B") !== -1) {
```
jobara: This will break if the string contains a mixture of BCI AV IDs and Blissary IDs. It isn't exactly clear from the docs that you can't do this. Interestingly having a mixture will result in an empty button on the palette but not display any errors in the errors section.

screenshot showing the palette preview with an empty button but no errors

docs/PaletteJsonGenerator.md
```diff
@@ -0,0 +1,73 @@
...
+SVG:B220;B99:SVGLABEL:children:LABEL
+```
+
+### SVG Builder String Syntax
+
+SVG builder strings take the form `SVG:<components>:SVG`.  Components are BCI AV IDs (plain numbers) or
+Blissary IDs (numbers prefixed with `B`).
+
+For full details on separators, relative kerning, letter syntax, and visual examples, expand the
+**"How to specify matches"** section within the running app.
```
jobara: This probably makes sense to include in the docs as well, and/or have a link to where it is in the running app.

docs/GenerateBlissSymbolExplanations.md
```diff
@@ -0,0 +1,181 @@
...
+
+## Steps
+
+### Step 1 — Download the input file
+
+Download the source dictionary from
+[Google Drive](https://drive.google.com/file/d/1UyhCDYeEe-OzgM7aWWCpBoEog9H9D5RL/view?usp=drive_link).
+
+Save it locally, for example at `scripts/data/bliss_dictionary_<date>.json`. When a new symbol set is
+released, an updated version of this file will be shared.
```
jobara, line 28: We can also just create a new repo just for this file and create an npm package for it with the appropriate license. Then we can version it and etc.

docs/GenerateBlissSymbolExplanations.md
```diff
@@ -0,0 +1,181 @@
...
+  public/data/bliss_symbol_explanations.json
+```
+
+Example:
+
+```bash
+node scripts/generate_bliss_symbol_explanations.js \
+  scripts/data/bliss_dictionary_20260513.json \
+  public/data/bliss_symbol_explanations.json
+```
```
jobara, line 44: This should be setup to run as an NPM script and can be triggered as a post install script there as well.

public/data/bliss_symbol_explanations.json
jobara, line 1: Was this file supposed to be committed to the repository?

public/palettes/a_an.json
jobara, line 1: Palettes should probably also include a description of what they are for. Some of the palette names do not seem to describe exactly what is inside of them. For example this one is called "a_an" but includes words like "against" and "forward".

scripts/generate_bliss_symbol_explanations.js
```diff
@@ -0,0 +1,228 @@
...
+ *     `code` -> decomposes the code into `composition`).
+ *    1.1 Fallback for missing `isChar`: defaults to `false`.
+ * 2. Composition Generation: Recursively decomposes non-character items 
+ *    (`isChar === false`) down to their base characters (`isChar === true`).
+ * 3. Delimiters: Inserts ";" before IDs that fall into the provided indicator 
+ *    ranges/lists, and "/" otherwise, dynamically flattening the array.
+ * 
+ * Notes but not show-stopper: after processing, a few items with missing fields:
+ * Warning: Items missing "explanation" value: 1 items: 6330
+ * Warning: missing "bciAvId": 9 items: 5844, 5995, 6435, 6436, 6437, 6438, 6439, 6440, 6441
```
jobara: I'm not exactly sure what this portion of the comment means.

scripts/generate_bliss_symbol_explanations.js
```diff
@@ -0,0 +1,228 @@
...
+  rawData = fs.readFileSync(inputFileName, "utf8");
+} catch {
+  console.error(`Error: Failed to read "${inputFileName}". Make sure the file exists.`);
+  process.exit(1);
+}
+
+// JSON.parse always returns any. The JSDoc cast on the declaration line doesn't
+// fix the linting error "@typescript-eslint/no-unsafe-assignment". So disable
+// that specific linting rule for this line.
+// eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
```
jobara: As this file is intended to be run directly from Node, it probably shouldn't be written in typescript and as such the typescript checks should probably not include this file.

scripts/generate_bliss_symbol_explanations.js
```diff
@@ -0,0 +1,228 @@
...
+const parsed = /** @type {{ data: BlissItem[] }} */ (JSON.parse(rawData));
```
jobara: Should this be wrapped in a try/catch block for cases where the file isn't JSON or it is malformed. In case you want to supply some helpful information.

scripts/generate_bliss_symbol_explanations.js
```diff
@@ -0,0 +1,228 @@
...
+/** @type {BlissItem[]} */
+const data = parsed.data;
+
+// Pass 1: Build a Lookup Map by Top-Level `id`
+/** @type {Map<number, BlissItem>} */
+const topLevelIdMap = new Map();
+data.forEach(item => {
+  // Default isChar if missing
+  if (!Object.prototype.hasOwnProperty.call(item, "isChar") || item.isChar === null) {
```
jobara: Can this be written like:


```suggestion
  if (!item.hasOwnProperty("isChar") || item.isChar === null) {
```

scripts/generate_bliss_symbol_explanations.js
```diff
@@ -0,0 +1,228 @@
...
+    item.isChar = false;
```
jobara: This is kind of a side effect, at least based on the comment on line 80 "Pass 1: Build a Lookup Map by Top-Level `id`"

scripts/generate_bliss_symbol_explanations.js
```diff
@@ -0,0 +1,228 @@
...
+  topLevelIdMap.set(Number(item.id), item);
+});
+
+// Pass 2: Transformation & Code Decomposition
+/**
+ * @param {BlissItem} item
+ * @param {Set<number>} visited
+ * @returns {(string | number)[]}
+ */
+function decompose(item, visited) {
```
jobara: I'm still trying to wrap my head around this function, but I'm wondering if this visited set would prevent a word such as [carnivore](https://blissary.com/dictionary/?q=B5649) from being fully decomposed? I'm also not sure if it needs to be fully decomposed, or if maybe it needs all possible decompositions. Such that there is an array of compositions. 

For example with carnivore you may have a set like:

- [B5649](https://blissary.com/dictionary/?q=B5649)
- [B122](https://blissary.com/dictionary/?q=B122)/[B1992](https://blissary.com/dictionary/?q=B1992)
- [B122](https://blissary.com/dictionary/?q=B122)/[B329](https://blissary.com/dictionary/?q=B329)/[B122](https://blissary.com/dictionary/?q=B122)

scripts/generate_bliss_symbol_explanations.js
```diff
@@ -0,0 +1,228 @@
...
+      const refId = parseInt(match[1], 10);
+      const refItem = topLevelIdMap.get(refId);
+
+      // Verify item exists and has a valid ID reference
+      if (refItem) {
+        // Prevent infinite loops
+        if (visited.has(refId)) {
+          reports.circularDependency.add(`Error: Circular dependency detected involving ID ${refId}.`);
+          continue;
+        }
```
jobara: Should the item be skipped if it's been seen already? Won't the decomposition be missing a part?

scripts/generate_bliss_symbol_explanations.js
```diff
@@ -0,0 +1,228 @@
...
+
+// Pass 3: Write Output File
+try {
+  fs.writeFileSync(outputFileName, JSON.stringify(outputData, null, 2), "utf8");
+} catch {
+  console.error(`Error: Failed to write to "${outputFileName}". Check directory permissions.`);
+  process.exit(1);
+}
+
+// 7. Output Reports
```
jobara: I think this should be 4, but not sure. Also the other's are called "Pass".

jobara: Perhaps all of these sections should be separated into functions.

scripts/generate_bliss_symbol_explanations.js
```diff
@@ -0,0 +1,228 @@
...
+console.log("\n=== Processing Report ===");
+console.log(`Report: Successfully processed ${data.length} records into ${outputFileName}\n`);
```
jobara: It would probably be nice to have a way to suppresses the outputs, maybe only have them displayed in a verbose mode.

src/client/ActionBmwCodeCell.test.ts
```diff
@@ -59,6 +59,10 @@ describe("ActionBmwCodeDell render tests", (): void => {
 
     // Check disabled state (should be enabled)
     expect(button.getAttribute("disabled")).toBe(null);
+
+    // Check that SVG is rendered (composition field must be set; bciAvId would leave it undefined)
+    const svgElement = button.querySelector("svg");
+    expect(svgElement).not.toBe(null);
```
jobara, line 66: It might be nice if the SVG contained the B ID and/or bciAvID if it is an existing Bliss character or word. Perhaps using a [data attribute](https://developer.mozilla.org/en-US/docs/Web/SVG/Reference/Attribute/data-*). @hlridge any thoughts on that?

If that were the case you could check for the value of that data attribute to make sure it's the correct one.

src/client/ActionGlossSearchCell.test.ts
```diff
@@ -83,6 +87,11 @@ describe("ActionGlossSearchCell render tests", (): void => {
     expect(inputElement).toBeVisible();
     expect(inputElement).toBeValid();
     expect(inputElement).toHaveValue(TEXT_SEARCH_PROPOSED_GLOSS);
+
+    const spanEl = document.querySelector(".actionGlossSearchCell span");
+    expect(spanEl).not.toBeNull();
+    // composition=3274 (number) → compositionToBstr(3274) = "B3274"
```
jobara: If this comment is meaningful it may need some more description around it, otherwise it should probably be removed.

src/client/ActionSearchGloss.test.ts
```diff
@@ -75,6 +75,6 @@ describe("ActionSearchGloss component behavior", () => {
     // When valid value is submitted
     await user.type(searchInput, "fish");
     await user.click(submitButton);
-    expect(await screen.findAllByText(/14188/i)).not.toHaveLength(0);
+    expect(await screen.findAllByText(/^fish: /i)).not.toHaveLength(0);
```
jobara, line 79: Did this not have a label before?

src/client/GlobalUtils.ts
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright 2024 Inclusive Design Research Centre, OCAD University
+ * Copyright 2024-2026 Inclusive Design Research Centre, OCAD University
```
jobara: You might want to switch to using an authors file like we did in Infusion so that you can have a standard statement in all the files and just update the dates in a single file.

src/client/SvgUtils.ts
```diff
@@ -10,390 +10,278 @@
...
+ * Handled Tokens:
+ * - Symbol IDs: "B" followed by digits (e.g., "B123"). These are parsed into numbers.
+ * - Kerning Codes: "AK:" or "RK:" followed by digits/negatives (e.g., "AK:-10"). Kept as strings.
+ * - Letter Codes: "X" followed by any Unicode letter (e.g., "XA", "Xα"). Kept as strings.
+ * - Separators: "/", "//", ";", and ";;". Kept as strings.
+ * 
+ * Validation:
+ * The string must be entirely composed of valid tokens. If the string is empty, 
+ * or if it contains any unrecognized garbage characters (e.g., "asdffr;B12"), 
+ * the function returns an empty array.
```
jobara, line 52: I wonder if in this case you should return the original string so that someone could see what the issue was.  Or perhaps also include the original composition string.

jobara (changes requested):

apps/palette-generator/index.html
cindyli: I agree.

apps/palette-generator/paletteJsonGenerator.ts
cindyli: The type definition for `bciAvId` is `bciAvId?: number,`. The optional flag implies the type is `number | undefined`. The fallback to `undefined` instead of `null` is to comply with this type.

src/client/ActionBmwCodeCell.test.ts
hlridge: @jobara are you thinking "you could check" as "a user can check", or how are you thinking this should be checked? Programmatically you could already check this with the builder.toString() function. Nevertheless, I will take your suggestion and think about it for future versions of Bliss SVG Builder. Thank you.

scripts/generate_bliss_symbol_explanations.js
cindyli: Using `item.hasOwnProperty` will cause linting error `no-prototype-builtins`. [Eslint site](https://eslint.org/docs/latest/rules/no-prototype-builtins) explains why `item.hasOwnProperty` is unsafe to use.

public/palettes/a_an.json
cindyli: Good point. The name "a_an" doesn’t describe the contents of that palette.

In this case, "a_an" is just an internal key defined in `public/palettes/palette_file_map.json`. It maps to a palette file. The name is somewhat arbitrary. It was likely picked because this sub-palette is opened from the “a, an” symbol cell. Since palettes can be reused and attached in multiple places, the key isn’t tied to the contents. So I'm not sure what information should be provided in the description. Please let me know if you have thoughts. 

For now, I’ve updated `docs/ClientDeveloperDoc.md` to explain how the map file works and how keys like "a_an" are used.

cindyli:
@jobara I've addressed all your comments except moving `public/data/bliss_symbol_explanations.json` into a new repo. Ready for another review. Thanks.

scripts/generate_bliss_symbol_explanations.js
jobara: I don't think there should be any prototype data though, as this is a custom object. However, I suppose because it's parsed something could be added to the original file that we don't know about.

See: https://eslint.org/docs/latest/rules/no-prototype-builtins#when-not-to-use-it

PR #135, opened 2026-06-06
Chore: remove RAG support (resolves #132)

cindyli:
# Pull Request Checklist

* [X] This pull request has been tested by running `npm run lint` without errors
* [X] This pull request has been tested by running `npm run typecheck` without errors
* [X] This pull request has been tested by running `npm run test` without errors
* [ ] This pull request has been tested by running `npm run dev` without errors
* [X] This pull request has been built by running `npm start` without errors
* [X] This isn't a duplicate of an existing pull request

## Description

This PR removes the RAG support and the entire server side because the server only provides RAG API and a /health check endpoint.

This PR is based on https://github.com/inclusive-design/adaptive-palette/pull/126 that should be reviewed and merged first.

Use this [compare URL](https://github.com/cindyli/adaptive-palette/compare/feat/implement-new-symbol-ids...chore/remove-rag-support) to review the changes between this PR and https://github.com/inclusive-design/adaptive-palette/pull/126.

## Steps to test

1. Run `npm run lint`, `npm run typecheck`, `npm run test` to make sure they pass.
2. Review the code to make sure RAG and server side have been removed.

**Expected behavior:** See above.

PR #155, opened 2026-07-20
feat: Update the symbol label when an indicator is applied (resolves #154)

### LINKED ISSUE #154: Feature: Update the symbol label when an indicator is applied

When an indicator is applied to a symbol, the symbol’s label does not update accordingly. For example, applying a past tense indicator to the symbol “to eat” still displays the label “to eat” instead of the correct form “ate”.

**Fallback tiers**

1.  Maintain a lookup table for common symbol + indicator combinations (e.g., verb tense transformations for action symbols).
   Example: "to eat" + past → "ate", "to go" + past → "went"
2. If a combination is not found in the lookup table, query a language model to generate the correct label dynamically.
3. If the language model is unavailable, keep the original label as a fallback.

### Additional context or notes

* Since querying a local model via Ollama can take quite long or Ollama may not be available on all devices, add a configuration option to:
  * Enable/disable language model fallback
  * Specify which model to use
* The lookup table should cover high-frequency verbs and common transformations to minimize reliance on the language model.
* Consider caching model-generated results to improve performance over time.

cindyli:
# Pull Request Checklist

## Description

This PR updates the symbol label when an indicator is applied.

The implementation is described in the documentation [`docs/IndicatorLabelLookup.md`](https://github.com/cindyli/adaptive-palette/blob/feat/new-labels-with-indicator/docs/IndicatorLabelLookup.md).

This PR is based off https://github.com/inclusive-design/adaptive-palette/pull/135 that should be reviewed and merged first.

Use this [compare URL](https://github.com/cindyli/adaptive-palette/compare/chore/remove-rag-support...feat/new-labels-with-indicator) to review the changes between this PR and https://github.com/inclusive-design/adaptive-palette/pull/135.

## Steps to test

1. Run `npm start` to start the application
2. Access in a browswer via `http://localhost:3000/`
3. Try symbol+indicator pairs to make sure the label changes automatically for pairs defined in [the lookup table](https://github.com/cindyli/adaptive-palette/blob/feat/new-labels-with-indicator/public/data/new_labels_with_indicator.json)
4. The fallback to Ollama model query is turned off by default. Turn it on if Ollama is installed on your testing computer. The setting is at this line in the config [file](https://github.com/cindyli/adaptive-palette/blob/feat/new-labels-with-indicator/public/config.json#L3). Also make sure the model config in the next line is an Ollama model you installed.
5. After changing the config file, re-run `npm start` to restart the app for the new config to be loaded when the app starts.
6. Choose a symbol+label combination that is not defined in the lookup table, wait around one minute for the label to be returned back by the Ollama query.
7. If the Ollama fallback is turned off or it's turned on but Ollama is unavailable, the label should remain unchanged.


hlridge (changes requested):
I scoped this to the changes specific to #155, the commits layered on top of #135 (`chore/remove-rag-support`) that it stacks on. #135's own changes are out of scope here since they belong to that PR's review. As your description notes, #135 should be reviewed and merged first; agreed, and I'm flagging it here so it stays visible at merge time.

Short version: the feature is well-built where it counts. The three-tier resolution order is correct, every runtime failure path (missing data file, bad config, unreachable Ollama, no dictionary id) degrades to "leave the label unchanged" rather than throwing, and the central promise that the glyph never blocks on the lookup genuinely holds. `ActionIndicatorCell` updates the composition and reassigns the signal synchronously before it `await`s `getNewLabel`, and the post-await `stillCurrent` check correctly drops a stale result if the symbol changed while the query was in flight. Nice work.

Two things I'd like fixed before merge, one design question about the live-query fallback, and a set of robustness items that don't affect the shipped table but are worth closing before the next table rebuild.

## Must fix

1. **Doc points at a file that doesn't exist.** `docs/IndicatorLabelLookup.md` (the "File locations" table) lists `src/client/IndicatorLabels.ts`, but the module is `src/client/IndicatorLabelsUtils.ts`. One-line fix.

2. **~57 MB of regeneratable build intermediates are committed, and nothing ignores them.**
   - `scripts/new_labels_with_indicator/data/new_labels_with_indicator_prompts.jsonl` (~28 MB) is deterministic input, regenerable from committed sources in seconds.
   - `scripts/new_labels_with_indicator/data/new_labels_with_indicator.jsonl` (~29 MB) is the raw LLM output, dead once the final JSON exists.
   - `scripts/new_labels_with_indicator/data/build_error.log` (3,437 lines) isn't produced by any script in this PR. It's `"Duplicate composition key ... overwriting previous label"` warnings from a different tool, so anyone opening it for the build's error history is chasing the wrong thing.

   `public/data/new_labels_with_indicator.json` (~3 MB) is the actual runtime asset and should stay. I'd drop the other three and add `scripts/new_labels_with_indicator/data/` to `.gitignore` (this PR doesn't touch `.gitignore` today). One caveat: a plain `rm` still leaves the blobs in the branch history, so we want to make sure they don't reach `main` (squash merge, or a history rewrite of the branch).

## One design question: the live-query fallback

The pregenerated table covers the shipped vocabulary. The second tier is a live model query (Ollama today, off by default), and per your own test steps it's roughly a one-minute round trip per label. Since a fair bit of this PR exists only to serve that tier, it's worth settling its future before treating the items under it as work:

- **Is this fallback meant to stay and eventually be switched on?** If yes, its cache wants a fix first: an empty model response is cached for the whole session and never retried, a thrown error is never cached so every click re-pays the full timeout, and because the cache stores resolved values set only after the await, a quick double-click fires two concurrent queries for one action. Caching a `Promise` stored before the await, with one policy for empty vs thrown, closes all three. These apply to whatever live model backs the tier, not only Ollama.
- **If it is not planned for active use, dropping the tier may be cleaner** than fixing it, and it removes the cache, the prompt-building, the model config, and the smaller items below from this PR.

I'm not assuming which. Three items hinge on the same answer, parked here rather than listed as separate to-dos: the label-only cache-key collision (with no `userSelectedSymbolId`, the key is `${baseLabel ?? label}_${indicatorId}`, so two hand-built compositions sharing a label collide), the builder-vs-runtime model mismatch (`gemma-4-31B-it` offline vs `gemma4:12b` in `config.json`), and the missing tests for the cache, the throw path, and `buildOllamaPrompt` returning undefined.

## Should fix (before the next table rebuild)
- **`build_final_labels_with_indicator.js` doesn't guard an empty or missing `newLabel`.** A future rebuild that emits `newLabel: ""` would ship `"{id}": ""`, and the client returns that `""` (it is `!== undefined`), so it skips the fallback and the "leave unchanged" path and blanks the label. A missing `newLabel` silently drops the entry instead. The shipped table has zero empty labels, so this is latent, but worth a hard check plus `process.exit(1)`. (The duplicate-`targetId` guard right above it is correct.)
- **`run_new_labels_with_indicator.py`:**
  - Resume appends in `"a"` mode after a possibly-partial trailing line, which can concatenate two rows into one unparseable line and brick the next build. Write via temp file plus rename, or truncate to the last newline on startup.
  - Generation is non-deterministic (no `do_sample=False` / seed), so a rebuild differs from the committed table in unchanged pairs purely from sampling drift, which makes future diffs hard to read.
  - The script exits 0 even when per-row errors occurred, so a short table looks like success and Slurm's FAIL mail never fires. `sys.exit(1)` if any errors.

## Nits

<details>
<summary>Expand</summary>

- Stale JSDoc in `IndicatorLabelsUtils.ts`: `getNewLabel` / `buildOllamaPrompt` document a `@param {SymbolEncodingType} payload` that no longer exists (real params are `userSelectedSymbolId, label, baseLabel, indicatorId`), and the cache-key description is out of date.
- `ActionIndicatorCell.ts`: the modifier-limitation comment says `baseLabel` is "derived from the bare dictionary word," but `baseLabel` actually carries the modifier ("big walk"); only the table's `newLabel` is bare. Also the success branch re-sets `baseLabel` to the value it already had before the await, so that line can go.
- `index.d.ts` types `indicatorInfo?: number[]`, but every write sets exactly one element and the engine reads a scalar. Either use a scalar or note the multi-indicator intent.
- `ActionRemoveIndicatorCell.ts` slice assumes a `;` precedes the indicator. Only reachable through malformed hand-authored palette JSON, but a `indicatorIndex > 0 && arr[indicatorIndex - 1] === ";"` guard is cheap.
- `ActionCodeCell.ts` / `ActionGlossSearchCell.ts` still set `"id": props.id` on the payload, a field the type no longer declares and that the modifier cells already dropped. Dead data.
- `job_run_new_labels_with_indicator.sh`: the license header URL points at `inclusive-design/baby-bliss-bot` (looks copied from `test_slm.py`), the output path doesn't match where the builder and doc expect it, `pip install --upgrade pip` runs before the venv is created, and `module load StdEnv/2023` appears twice. Also worth confirming the pinned `torch` / `transformers` wheel tags resolve on the Alliance cluster.

</details>

## Tests

Happy paths are well covered (tier-1 hit and miss, Ollama prompt shape and model, positive cache, swap re-derives from `baseLabel`). The gaps I'd most want closed:

- The "glyph renders before the label resolves" invariant: a deferred `getNewLabel` mock, asserting the composition is already updated pre-resolution.
- The `stillCurrent === false` race: click A, click B before A resolves, resolve A, assert A's result is dropped.
- The modifier + indicator interaction: the documented limitation is entirely unasserted, so a future "fix" would land with no guard to catch a regression.
- `ollamaCache` isn't resettable between tests, which is an isolation leak waiting for the first key reuse.

Overall this is close. Fix the doc path and the committed intermediates and I'm happy; the rest can be follow-ups. The main thing I'd want your read on is the live-query fallback question above.

cindyli:
@hlridge Thanks so much for the review. I've addressed them all except this one:

> The label-only cache-key collision (with no `userSelectedSymbolId`, the key is `${baseLabel ?? label}_${indicatorId}`.

When querying a model for the new label when an indicator is applied, the prompt used is based on the symbol label rather than the symbol. I think it makes sense to have the cache key based on `symbolLabel + indicatorId` rather than including `symbolId` as well.

This PR is ready for another review. Thanks.

hlridge (approved):
Thanks for the thorough pass. Agreed on the cache key: since the prompt in the no-`userSelectedSymbolId` branch is built solely from `baseLabel ?? label`, two compositions sharing that label produce the same prompt and should resolve to the same answer, so keying on label + indicator is correct, not a collision. The key mirrors the prompt inputs in both branches.

Everything else I raised is addressed, and the indicator-removal modifier handling went further than I asked. Two trivial leftovers, both non-blocking: the SLURM script still runs `pip install --upgrade pip` before the venv is created, and loads `StdEnv/2023` twice. LGTM.

cindyli:
Thanks for your approval, @hlridge.

Regarding the suggestion below, the same set of commands being called twice is intentional and not a bug. This is required for the job script on Alliance. The first call makes sure the OS environment is properly configured for setting up the virtual environment. The second call then configures the environment inside the virtual environment itself.

> the SLURM script still runs `pip install --upgrade pip` before the venv is created, and loads `StdEnv/2023` twice.

cindyli:
Thanks for your approval, @hlridge.

Regarding the suggestion below, the same set of commands being called twice is intentional and not a bug. This is required for the job script on Alliance. The first call makes sure the OS environment is properly configured for setting up the virtual environment. The second call then configures the environment inside the virtual environment itself.

> the SLURM script still runs `pip install --upgrade pip` before the venv is created, and loads `StdEnv/2023` twice.

PR #163, opened 2026-07-27
feat: Simplify the feature that translates a telegraphic message to a full sentence (Resolves #162)

### LINKED ISSUE #162: Feature: Simplify the feature that translates a telegraphic message to a full sentence

The adaptive palette currently displays "Prompts" and "Ask AI" sections at the top of the page. These features allow users to manually adjust prompts and select a language model to convert their telegraphic input into a complete sentence. This approach requires users to understand how to construct effective prompts, and can choose a suitable model for this type of task. In practice, this creates unnecessary complexity especially for users who may not have expertise in prompt engineering or model selection.

Recently we've done testing to identify better prompts and models for this translation task. Based on these findings, we can simplify the user experience by removing the need for manual configuration.

The solution includes:

* Remove the "Prompts" and "Ask AI" sections from the interface
* Provide a button that automatically performs the translation using pre-configured model and prompt
* Add a section in `config.json` to set the model name and system/user prompts

cindyli:
# Pull Request Checklist

* [X] This pull request has been tested by running `npm start` without errors

## Description

This PR simplify the process that translates a telegraphic message to a full sentence. See the [issue](https://github.com/inclusive-design/adaptive-palette/issues/162) for details.

## Steps to test

1. Run `npm test`, `npm run lint`, `npm run typecheck` to make sure they don't report errors.
2. Run `npm start`, check the web app and the user workflow to make sure it works in a reasonable way.
3. Review the code in this branch.

**Expected behavior:** The app should word as expected and code is clean


hlridge (changes requested):
## Review of #163

Nice piece of work. The state machine in `TelegraphicTranslationState.ts` is the strongest part: the
stale-reply guard, the abort-versus-error distinction, and the `finally` block that refuses to
unregister a successor controller are all things that usually get discovered in production rather
than written up front, and each one has a test naming the scenario it protects. Extracting
`CellTypeRegistry.ts` out of `GlobalData.ts` and documenting the import rule is a good side effect.

The `structuredClone` snapshot in `ee36e25` is a good catch. The editing cells mutate `payloads` and
`modifierInfo` in place and then publish a fresh wrapper, so a snapshot of the wrapper was never a
snapshot of the message, and the revert path would have restored already-mutated arrays. The test
that reproduces it by pushing into `payloads[0].modifierInfo` before republishing is exactly the
right shape.

I verified the checklist locally on `5595b6e`: `tsc --noEmit` clean, `markdownlint-cli2` 0 issues,
`eslint` clean, and 219 tests over 33 files pass in chromium. Firefox and webkit were not run, so those two instances are unverified from my side. The
`:has()` rule below is one I would want confirmed on webkit in particular.

### Accessibility: the live region is hidden while it is idle

`SentenceChoices.scss:40`

```scss
.sentenceStatus:empty {
  display: none;
}
```

The rest of this design is explicitly built on the region already being present so the screen reader
is watching it: the CSS comment above says so, `SentenceChoices.ts:39` says so, and two tests assert
that the post-update element is the same node
(`"the working message lands in the live region that was already there"`). But `display: none` takes
the element out of the accessibility tree, so while idle there is nothing for the screen reader to
watch. The text insertion and the switch back to `display: block` land in the same commit, which is
exactly the "region appears with the text already in it" case that NVDA and JAWS routinely miss.

The tests pass because they check DOM identity, and `toBeVisible()` is only asserted after the text
has arrived, by which point the rule no longer matches.

Suggest the visually hidden idiom instead, so the element is always rendered:

```scss
.sentenceStatus:empty {
  clip-path: inset(50%);
  height: 1px;
  margin: 0;
  overflow: hidden;
  padding: 0;
  white-space: nowrap;
  width: 1px;
}
```

Worth a manual pass with NVDA or VoiceOver either way, since this is the one part of the feature no
automated test can confirm.

### The column takeover is layout logic living in CSS

`ContentEncoding.scss:16-18`

```scss
.paletteContainer:not(:has(.btn-makeSentence)) .contentEncodingArea {
  grid-column: 1 / 12 !important;
}
```

The behaviour is right and I like that the test asserts real computed style rather than a class name.
Three things about the mechanism though.

**It is a second source of truth for the same layout.** The content area's span comes from
`input_area.json` (`columnStart: 1, columnSpan: 10`), which `ContentEncoding.ts:134` turns into an
inline `style`. The `1 / 12` here is that value plus one, hardcoded, in a different file and a
different language. Edit the palette JSON and this silently goes stale, and because it is
`!important` it will win over the JSON while being wrong.

**The `!important` is a symptom rather than a choice.** It is needed only because the span arrives as
an inline style. Deciding the span where the span is already decided removes the need for all three
of `!important`, `:has()`, and the class dependency. `CommandMakeSentence.ts:47-48` already computes
the governing condition:

```js
adaptivePaletteGlobals.LLMs.length === 0 || !adaptivePaletteGlobals.config.telegraphicTranslation
```

Having `ContentEncoding` widen its own `columnSpan` by one under that same condition puts both halves
of the decision in one place, with the JSON still the base value.

**The selector is global, not scoped to the input area.** `Palette.ts:79` renders every palette into
`.paletteContainer`, and `palette-generator.ts:139` renders through the same component. So in the
palette generator, any palette a user builds that contains a `ContentEncoding` cell and no
`CommandMakeSentence` cell gets its content area forced to `1 / 12 !important`, overriding the
`columnStart` and `columnSpan` that user just typed in. Narrow, but it is a preview that now lies.
`input_area.json` is the only shipped palette with a `ContentEncoding` cell, so scoping the selector
to `#inputArea` would cost nothing.

Also worth a note in `CommandMakeSentence.scss` that `.btn-makeSentence` is load-bearing in another
component's stylesheet. Renaming it today breaks the layout with no error anywhere.

### The documented config-error banner does not exist

`docs/TelegraphicMessageTranslation.md:159` promises:

> `config.telegraphicTranslation` is missing or invalid | A configuration error banner is displayed
> and the **Sentence** button is not shown.

The button does hide itself (`CommandMakeSentence.ts:47-50`), but nothing renders a banner.
`index.js:61-62` writes `aiStatus` only when `LLMs.length === 0`, and `NOT_CONFIGURED_MESSAGE` is
thrown only from `requestSentences`, which cannot be reached while the button is absent. So a typo in
`public/config.json` produces an app that silently has no Sentence button and no explanation, and
`NOT_CONFIGURED_MESSAGE` is unreachable outside its unit test.

Since a malformed section is a deployment mistake rather than a user state, I would wire the banner
rather than soften the doc: in `index.js`, when `LLMs.length > 0 &&
!adaptivePaletteGlobals.config.telegraphicTranslation`, write `NOT_CONFIGURED_MESSAGE` into
`aiStatus`.

### The doc now contradicts itself about the confirm dialog

The new paragraph at line 111 describes it correctly:

> If the input area is modified while a query is in progress or while sentences are displayed on the
> screen, a warning dialog will appear.

But the Edge Cases table two sections down still describes the behaviour from before that commit:

> line 163: Message edited while choices are visible | Existing choices are discarded because they no
> longer match the current message.
> line 164: Message edited or cleared while a request is in progress | The response is ignored when it
> arrives.

Both rows need the prompt folded in, and the "or cleared" half of 164 is worth splitting out, since
clearing through **Delete all** deliberately does not prompt (`clearMessageAndChoices` sets `idle`
first). A third row for the error state would round it out, now that editing clears it silently.

### Design question: `window.confirm` as the discard prompt

`TelegraphicTranslationState.ts:141-160`

Every other interaction in this feature is a large, styled, in-page button, which is the right call
for switch and eye-gaze users. The discard prompt is the exception: a native modal lives in browser
chrome, so it is outside the app's scanning surface, its targets are small and unstylable, its button
labels cannot be changed, and it blocks the main thread until answered. For the users this project
exists for, that prompt may simply be unreachable.

Two alternatives, in the order I would rank them:

1. **Non-blocking undo.** Apply the edit, discard the sentences, and show
   "Sentences discarded. Undo" as a button in the sentence area. No modal, no timing pressure, and
   the destructive action stays reversible. Probably the better fit for AAC: it does not make the
   user answer a question mid-thought.
2. **In-page confirm.** Two large buttons in the sentence area, matching the rest of the feature.

Either way the logic you have already written (snapshot, revert, abort) carries over almost
unchanged, and `WORKING_DISCARD_PROMPT` / `READY_DISCARD_PROMPT` stay as they are. Happy to be
overruled if the confirm was a deliberate call after testing with users.

### Smaller items

**`SentenceChoices.ts:24`** the placeholder reads `"None of above â€” let me type it"`, which is
missing a "the". I see `ee36e25` changed the doc to match the code here; I think it wants to go the
other way. The string is both the placeholder and the `aria-label`, so a screen reader speaks it.

**`TelegraphicTranslationState.ts:155`** the new error-clearing branch fires on any change to
`changeEncodingContents`, including a caret move that leaves the message identical. Harmless, and
arguably fine, but it is the one place in the effect that does not check whether the message actually
changed, so `"moving the caret does not ask"` has no counterpart covering that the error banner
survives a caret move. Worth deciding deliberately rather than by omission.

**`SentenceLog.ts:60-61`** the log composition sits outside the `try`:

```js
const entries = [
  ...readSentenceLog().filter((entry) => entry.telegraphicMessage !== record.telegraphicMessage),
```

`readSentenceLog` guards the parse, but `parsed as SentenceRecordType[]` does not validate elements,
so storage holding well-formed-but-wrong-shape JSON (`[null]`, `[1, 2]`) throws a `TypeError` out of
the click handler. The existing test covers unparseable JSON, not this. One-line fix in
`readSentenceLog`: `parsed.filter((entry) => typeof entry === "object" && entry !== null)`.

**`OllamaApi.ts`** the ~20 line comment explaining why `stream` cannot be passed as
`stream: streamResp` (TS2769, ollama-js#78) was deleted, but the `if`/`else` it justifies is still
there. That is the kind of comment that stops the next person "simplifying" it back into the
compile error. A two-line version pointing at the issue would be enough.

**`CommandMakeSentence.ts:25`** `MAKE_SENTENCE_LABEL = "Make a sentence"` is exported but only ever
read by `CommandMakeSentence.test.ts`; production takes its label from `input_area.json`, where it is
`"Sentence"`. A constant that only the test uses, and that does not match the shipped label, reads as
though it were the button's real name. Either use it as the component's default label or drop it and
let the test own its fixture string.

**`TelegraphicTranslationUtils.ts:64`** `parseSentences` strips `1.` and `1)` but not bullets, so a
model that answers with `- I am hungry` yields a sentence with a leading dash, which then gets
spoken. Cheap widening: `/^(?:\d+[.)]|[-*â€¢])\s*/`.

**`TelegraphicTranslationUtils.ts:30-37`** `pickModel` returns `LLMs[0]`, which is `undefined` typed
as `string` when the list is empty. It is guarded today by the `LLMs.length === 0` check in
`requestSentences`, but the guard lives in the caller while the function is exported and separately
tested. An early return or a line in the doc comment would pin the contract down.

**`GlobalData.ts`** `SYSTEM_PROMPTS_KEY` is no longer written, but anyone who ran an earlier build
keeps `"Telegraphic System Prompts"` in local storage indefinitely. A `removeItem` in
`initAdaptivePaletteGlobals` would tidy that up.

**Typos and comment nits**

- `TelegraphicTranslationState.ts:86` "Senteneces" -> "Sentences"
- `TelegraphicTranslationState.ts:137` trailing whitespace
- `SentenceChoices.ts:39` "annouce" -> "announce"
- `GlobalData.ts:84` "3.The" -> "3. The"; that block also opens with `/*` while carrying `@param` and
  `@returns`, and has trailing whitespace on several lines. The rest of the file uses `/**`
- `GlobalData.ts` `loadConfig`'s doc comment still says it validates only the `indicatorLabelLookup`
  section
- `TelegraphicTranslationState.test.ts:69` "Every the input message when a request is in flight asks
  the user to confirm the discard" looks like a sentence that lost a word
- `OllamaApi.test.ts:45` `describe("ollamaApi unit tests")` kept the old casing after the rename

### Diff scope

Two things inflate the diff without belonging to the feature, both harmless, noting them only so the
next reviewer does not go hunting for a reason:

- `GlobalUtils.ts` moved from a trailing `export { ... }` block to inline `export` keywords. Every
  symbol in that block was already exported, so the only functional change in the file is the new
  `normalizeComposition` (which is a genuinely good extraction from `ActionCodeCell`).
- `package-lock.json` carries an `npm update`. `package.json` is untouched, so no range decisions
  were made, but the lockfile already contains the bumps that renovate #164 (globals 17.8.0) and #166
  (@types/node 26.1.2) are proposing, so those two will need closing or rebasing after this merges.

### Checked and correct

For the record, things I went looking for and found in order:

- `"composition": 2351` is the internal id for "sentence, clause, phrase" (bciAvId 16760,
  `[368, "/", 502, "/", 414]`, "group of + word"), which is a good fit for the button, and bare-number
  compositions match the existing convention. The column shifts to 12-15 are consistent and
  `.paletteContainer` is an unconstrained grid, so nothing overflows.
- No orphaned references anywhere to the four deleted modules or to `ActionTextCell`, including in
  palette JSON, docs, demos, and `index.html`.
- `aria-controls="sentenceChoices"` resolves to a real element.
- Reverting inside the signals effect re-enters it once and converges, and
  `"a second refused edit is still put back"` covers the snapshot bookkeeping that makes it work.
- Keying the focus effect on the signal value's object identity means typing in the text box does not
  steal focus back to the first choice. Subtle, and right.
- Splitting the two config parsers means a broken `indicatorLabelLookup` section no longer discards a
  valid `telegraphicTranslation` section, which is a real improvement over the old all-or-nothing
  `disabled` object.
- Deduping the log by `telegraphicMessage` means single-sentence mode's `"auto"` record is replaced
  rather than duplicated when the user then taps the sentence.
- `aria-disabled` instead of `disabled` on the trigger, with the actual guard in `makeSentences`, so
  a click in the unavailable state is inert without the button losing focus.

cindyli:
Thanks for the review, @hlridge. Many good catches.

I've addressed all comments except `Design question: window.confirm as the discard prompt`. I agree `window.confirm` has accessibility issue. But on the other hand, I feel this is a temporary design and the interaction may change soon so I used this lowest cost solution. I will do a proper implementation when a more firmed design is in place.

hlridge (approved):
Re-reviewed at `beda18e`. Everything I raised is addressed, and a few of the fixes are better than
what I suggested. Approving.

On `window.confirm`: agreed, and thanks for saying so explicitly rather than just silently keeping
it. Shipping the cheap version of an interaction that is about to be redesigned is the right call.
One request, and it is a request rather than a condition: open a follow-up issue for it now, while
the reasoning is fresh. "Temporary" accessibility gaps in AAC software have a way of becoming
permanent ones once the surrounding design settles and nobody remembers the modal was a placeholder.
A one-line issue linking back to this thread would be enough.

Verified on `beda18e`: `tsc --noEmit` clean, `markdownlint-cli2` 0 issues, `eslint` clean, 220 tests
over 33 files pass in chromium. Firefox and webkit still unverified from my side.

Spot checks on the two fixes where a passing test would not have proven much:

- The `#inputArea` scoping resolves against the real app DOM, not just the fixture:
  `index.html:37-39` gives `#inputArea > #input_palette`, `index.js:52` renders the palette there,
  and `Palette.ts:78` emits `.paletteContainer` inside it. The updated test builds the same nesting
  rather than asserting a selector that only exists in the test.
- Hiding the button does not shrink the grid. `Palette.ts:79` derives
  `grid-template-columns` from the palette JSON, which still declares the cell at column 11, so the
  track stays and the widened content area fills it rather than leaving a gap.

Two fixes I would call improvements on what I asked for:

- Moving the empty-model guard into `pickModel` rather than leaving it in `requestSentences` puts the
  check where the contract lives, so the export is safe on its own terms. The `@throws` tag and the
  test that exercises it finish the job.
- `position: absolute` on the hidden live region. I left that out of my snippet and it matters:
  without it the 1px box still participates in the grid.

Three trivia, none worth another round. Fold them in if you touch the branch again, or leave them:

- `OllamaApi.ts:67-68` the two new comment lines end in a trailing space. Ironic given the same
  commit cleaned several of those up elsewhere. `eslint` does not enforce it here, so nothing fails.
- `OllamaApi.ts:66-70` the restored TS2769 note sits above `const request` with a blank line between
  it and the `if (streamResp)` at line 84 that it actually explains, so it reads as documenting the
  request object. Moving it down to just above the `if` would put it where the next person looks.
- `index.js:64` `else if (adaptivePaletteGlobals.LLMs.length > 0 && ...)` has a redundant first half
  inside the `else` of `LLMs.length === 0`. My fault, I wrote the condition that way in the review
  comment.

One non-blocking observation on the new placeholder, `"None fit? Type yours to help the tool learn"`:
it reads as a promise that typing improves the tool, but the records are inert local storage awaiting
future tuning, so nothing learns from them today. Not worth changing now, just worth knowing it is a
claim the app does not yet keep.

The `1 / 12 !important` pair is still two sources of truth for one layout, but scoping was one of the
two options I offered and it removes the part that actually misbehaved. Fine as it stands.

PR #169, opened 2026-07-31
feat: Move "SVG builder string" and "search by vocabulary" features into modal dialogs (resolves #168)

### LINKED ISSUE #168: Feature: Move "SVG builder string" and "search by vocabulary" features into modal dialogs

Currently, the palette page has two expandable sections above the input area: **"SVG builder string"** and **"Search by vocabulary"**. 

When expanded, these sections occupy a large vertical portion of the screen, pushing the main palette out of view. This causes UI shifting and requires unnecessary scrolling to get back to the primary tool.

Two changes:

1. Move these features out of the main vertical flow. They could be placed into modal dialogs so they don't consume primary vertical space on the palette page.
2. Allow users to enable/disable these sections in the settings based on the user's needs:
   * "Search by vocabulary" is for users who are able to type letters.
   * "SVG builder string is designed for developers.

cindyli:
# Pull Request Checklist

* [X] This pull request has been tested by running `npm start` without errors

## Description

This PR simplifies the UI above the input area to provide a cleaner user experience. 

**Changes made:**
* Replaced the bulky "SVG builder string" and "search by vocabulary" sections with two streamlined entry buttons.
* Configured default visibility settings based on the target audience (users vs. developers).

By default, the user-facing **Vocabulary** button is shown, and the developer-facing **SVG-Builder** button is hidden. Visibility can be toggled in `public/config.json` using the following properties:

```json
  "symbolSearch": { "show": true },
  "svgBuilderString": { "show": false }
```

## Steps to test

1. run `npm start`
2. Open the palette in a browser.
3. Adjust `public/config.json` when test these features.
4. Make sure they work properly.


hlridge (changes requested):
## Review of #169

Reviewed at `756ba79`. One defect worth fixing before merge and one test I would add. Everything else checked out, including the parts I expected to find problems in.

Building on the native `<dialog>` rather than hand-rolling was the right call, and the comments
explaining *why* at each trap (the UA `display: none` override, the `min-height: 0` flex chain, the
two WebKit divergences) are the kind that stop the next person undoing the fix by accident.

Verified: `tsc --noEmit` clean, `markdownlint-cli2` 0 issues, `eslint` clean, and 256 tests over 35 files
pass on all three browser instances (chromium, firefox, webkit).

### An empty label silently drops a word from the message

`ActionSearchGloss.ts:114` and `:127` use `labelDraft` raw:

```js
const payload: SymbolEncodingType = {
  label: labelDraft,
  ...
setStatus(`${labelDraft} added to message`);
```

The label field is pre-filled from the selection, but it is editable on purpose, so a user can clear
it and press **Add to message**. Two consequences:

1. The status region announces `" added to message"` with no subject, which a screen reader reads as
   a sentence missing its noun.
2. More seriously, the symbol goes into the message with `label: ""`, and
   `currentTelegraphicMessage()` builds the model's input as
   `payloads.map((payload) => payload.label).join(" ")`. An empty label contributes nothing but an
   extra space, so the word disappears from what the model is asked to translate while its symbol is
   still visibly sitting in the input area. The user sees a message the model never receives.

The sibling dialog already handles this. `ActionSvgEntryField.ts:87` does
`${labelString.trim() || "Symbol"}`, and has a test for it ("announces generically when no label was
given"). The search dialog is the user-facing one of the two, so it is the one that needs the guard
more.

Trimming would be worth doing at the same time, for the same reason: leading and trailing spaces
flow straight into the joined telegraphic message.

### Initial dialog focus has no test behind it

`ModalDialog.test.ts` covers focus *return* ("returns focus to whatever opened it"), which is the
harder half. Nothing covers where focus lands on open.

It currently lands in the right place, in the search field rather than on the `âœ•` dismiss button that
precedes it in DOM order, but only because two separate things hold: Preact reflects the bare
`autofocus` prop onto the content attribute, and the spec's dialog focusing steps prefer an
`autofocus` element over document order. Both are correct today. Neither would fail a test if a
Preact upgrade changed how that prop is applied, and the symptom would be quiet: every dialog opening
onto its close button, which for a switch or eye-gaze user is a scan cycle wasted on every use.

A test asserting `document.activeElement` is the search input after opening would pin it down.

### Notes, not defects

- `aria-pressed` on every result means each of up to 50 buttons announces "not pressed" during a
  scan. The comment at `GlossSearchResults.ts:31-34` weighs this against the `listbox` and
  `radiogroup` alternatives and picks per-result tab stops for switch users, which I agree with. Just
  naming the cost, since it is real and paid on every scan.
- The Label field sits outside the `<form>`, so Enter in it does nothing. Given **Add to message** is
  right beside it, this is arguable rather than wrong.

### Checked and correct

- `&[open] { display: flex }` rather than an unconditional `display`, which would have overridden the
  UA stylesheet's `dialog:not([open]) { display: none }` and left the dialog visible while closed.
- The scroll chain holds: `overflow: hidden` on the shell, exactly one `overflow-y: auto` descendant
  per dialog, and `min-height: 0` on every flex ancestor between them, so the footer cannot be pushed
  out of reach on a short or magnified viewport.
- Live regions are always present with a reserved `min-height` and are never `display: none`, and
  `#symbolEntryToolbar:empty` correctly notes why the same rule is safe there.
- Escaping the search term in `findSymbolByGloss` fixes a real crash: `(` or `[` previously reached
  `new RegExp` unescaped and threw. Covered by a test.
- Backquote no longer navigates the palette behind an open dialog.
- Each config section validates independently with a safe default, so `symbolSearch` and
  `svgBuilderString` cannot arrive undefined and take `SymbolEntryToolbar` down with them. The
  defaults are asymmetric in the right direction: an older config.json keeps symbol search and leaves
  the developer tool off.
- Both WebKit divergences are handled and explained: the opener captured on the trigger's click
  because WebKit does not focus a button on mouse click, and focus restored explicitly on close
  because WebKit does not return it.
- `useId` for the preview's label, since one component is shared by two dialogs.
- The two close controls have distinct accessible names.
- No orphaned references to `ActionGlossSearchCell`, `GlossSearchPalette`, or
  `findSymbolsContainingBciAvId`, and the stale `#SvgEntryDialog` / `#SearchGlossDialog` rules are
  gone from `index.scss`.
- `package-lock.json` moves are patch and minor only, `package.json` untouched, and no renovate PR is
  currently open to collide with it.

cindyli:
@hlridge thanks for the review. Your comments have been addressed. Ready for another check.

cindyli:
@hlridge, I added a couple of minor improvements in this PR:

1. String replacement that replaces the "LLM" with "model";
2. Updated the config file (`public/config.json`) by adding the system prompt and the user prompt. These prompts are used to query model with the new symbol label when an indicator is applied.

hlridge (approved):
Approved at `bad4350`. Both items from the last round are fixed. Checked: `tsc --noEmit`,
`markdownlint-cli2` and `eslint` clean, 272 tests over 35 files pass on chromium, firefox and webkit,
no `LLM` left in the tree.

One side effect you may not have meant, and I think it is an improvement: adding an unlabelled symbol
no longer triggers the discard warning, since the message those sentences came from is unchanged.

**Not blocking, `IndicatorLabelsUtils.ts:94`.** A builder string symbol can carry `label: ""` and has
no dictionary id, so applying an indicator to it builds the prompt from an empty word.
`renderPromptLines()` drops the `Word:` line, so all the model gets is the `Indicator:` line with no
word anywhere in it. It answers anyway, and the answer becomes that symbol's label. The cache key
collapses to `"_97"` as well, so every unlabelled symbol shares one answer. Both
close by returning `undefined` from `buildOllamaPrompt()` when the word is blank, landing on Tier 3
with the label unchanged. Predates this PR and the dialog ships hidden, so fix or file, your call.

Nits: `apps/ollama/ollama.html:3` has a `<title>` of "models" against its own `<h1>` "Models", quoted
lowercase again in `demos/Ollama Chat Service/README.md:70`, and
`demos/Ollama Chat Service/whatExpress.html:12` reads "a list of model's available".

